PR #2, opened 2024-10-25
Get working as a library

almarklein:
Copied some files over to make this a proper library, and make it actually installable / work.

See commit messages for what this does.


almarklein:
@Korijn I'm moving fast until I get to the point where we can merge in https://github.com/pygfx/wgpu-py/pull/618

Most of this is copying stuff over with no/minimal changes.

PR #4, opened 2024-10-25
Add CI

almarklein:
Everything should be in place. Will fail for the time being.


almarklein:
Linting and dist building is green.

Issue #1, opened 2024-10-24
Setting things up


almarklein:
## Setting up

* [x] Use `git-filter-repo` to clone wgpu-py, but only the gui stuff (`wgpu/gui/*.py` and `tests/test_gui_*.py`).
* [x] Create repo at Github 👋  and push the reduced clone.
* [ ] Move `wgpu/gui` to `rendercanvas`.
* [ ] Populate repo with readme, gitginore, pyproject etc.
* [ ] Make it work with `from rendercanvas import WgpuCanvas`
* [ ] Add the examples.

## Moving forward 

* [ ] Add docs enough so we can do the below.
* [ ] Open https://github.com/pygfx/wgpu-py/pull/618 as a pr here, and get it merged.
* [ ] Rename `WgpuCanvas` to `RenderCanvas`.
* [ ] Get rid of other references of wgpu.  
* [ ] Make docs work and clean them up.
* [ ] Rename and clean up examples.

## More short term work

* [ ] Handle sigint.
* [ ] Animate events.
* [ ] Allow `draw_func` to be an async function.
* [ ] Can we move present() out of CanvasContext()?
* [ ] Register the draw function as a draw event handler instead of with `request_draw()`?
* [ ] Allow requesting a draw by setting a field in the event object?
* [ ] Tracking statistics (track time spent on different things). Plus visualizing them. Maybe later.

## Adoption

* [ ] Move issues from wgpu-py repo to here.
* [ ] Add `rendercanvas` version to diagnostics of wgpu-py.
* [ ] Use rendercanvas downstream.
* [ ] Deprecate `wgpu.gui`.
* [ ] Remove `wgpu.gui`.

almarklein:
<img width="300" alt="Screenshot 2024-10-25 at 11 18 49" src="https://github.com/user-attachments/assets/2f41b823-afbd-40d8-959a-38b9e7327af5">
<img width="300" alt="Screenshot 2024-10-25 at 11 20 29" src="https://github.com/user-attachments/assets/612ede07-254a-4956-a5d4-dfceb8c81a7e">

almarklein:
This issue was meant to track progress in the first weeks/months after spinning rendercanvas out of `wgpu.gui`. I created separate issues for remaining open points. Closing.

PR #5, opened 2024-10-25
Add docs

almarklein:
Copied docs over from wgpu-py, removed some stuff, added a tiny bit. Kept most of it as-is, because we have a big pr to merge in. I'm making a note in #1 to cleanup the docs in the end.


almarklein:
RTD also builds on each pr.

PR #17, opened 2024-11-12
Add backends for all qt libs

almarklein:
* [x] Adds shims so one can do `from rendercanvas.pyside6 import RenderCanvas, loop`.
* [x] Fix a few problems with PyQt6 😅 
* [x] Document this.  

When selecting the qt backend via these modules, it can be that `rendercanvas.qt` has already been imported with another qt-lib. That's why we also do a check to make sure it's actually the intended lib being used.

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -40,16 +40,17 @@ but you can replace ``from rendercanvas.auto`` with ``from rendercanvas.glfw`` t
 Support for Qt
 --------------
 
-RenderCanvas has support for PyQt5, PyQt6, PySide2 and PySide6. It detects what
-qt library you are using by looking what module has been imported.
+RenderCanvas has support for PyQt5, PyQt6, PySide2 and PySide6.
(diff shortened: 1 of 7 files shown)


hmaarrfk:
Out of curiosity is using qtpy that bad?

almarklein:
> Out of curiosity is using qtpy that bad?

I think qtpy solves a different problem. We don't really need it because the API that we use of Qt is so small that dealing with the different Qt libs is easily done:

https://github.com/pygfx/rendercanvas/blob/44c2c268f941999efa54f838b8c49f9d968561d7/rendercanvas/qt.py#L33-L48

This PR is really to scratch a little itch that in order to run an example in qt (e.g. for testing stuff) we had to write
```py
import PySide  # this line is needed so rendercanvas.qt knows what lib to use
from rendercanvas.qt import ...
```

A small annoyance, to be fair, but it felt inconsistent. With this pr I can just replace the  "auto" in `rendercanvas.auto` to "pyside6". 

For users using the rendercanvas in a Qt app, I'd recommend to just keep using `from rendercanvas.qt ...`.

PR #21, opened 2024-11-14
Refactor inter-context-canvas api

### LINKED ISSUE #20: Scope: who can render to our canvas?
### Intro

Our `canvas.get_context()` always returns a `WgpuCanvasContext`. I think we can release this a bit, opening options for other tools that want to render to a canvas, without adding much complexity on our end. I did some thinking and experimenting today. In this issue I try to stick to the high-level ideas; I have a PR underway with implementation.

### Rendering vs presenting

Let's get some terms straight. The context helps the user-code to *render* to a canvas. The context *presents* the result to the canvas. From the p.o.v. of RenderCanvas, we don't care what API the context provides the user with. We only care that the context presents the result in a way that the canvas supports.
```                                  
            ┌─────────┐      ┌────────┐ 
            │         │      │        │ 
            │ Context │      │ Canvas │ 
            │         │      │        │ 
user        └─────────┘      └────────┘ 
    ───────►          ──────►           
     render            present          
```
### Some thoughts / observations / statements

* I think `rendercanvas` should be mostly agnostic about wgpu.
* The `get_context()` in WebGPU is because `get_context()` [already exists](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) for webgl, 2D api, etc. Strictly speaking I don't mind using a completely different API here, but in the end I think this mechanic/abstraction actually works well for our case.
* In the case of wgpu, the context implementation is provided by the wgpu library. That's where the logic is that can turn a window id into a surface id, and configure it. In other words `rendercanvas` cannot provide the context itself, but must load it from another library.
* I don't want to make an effort to support OpenGL.
* Canvases generally support either rendering to screen, or rendering to an image, sometimes both. Some GUI libs support OpenGL, but see the previous point

### Proposal

To get a canvas:

* Use `canvas.get_contex(context_type)` to allow selecting a context:
  * E.g. `canvas.get_contex("wgpu")` or `canvas.get_contex("something_else")`
* Define a hook so that `context_type` can be  "some.module.name", and if that module contains `context_from_rendercanvas_hook()`, we call it to get the context.
* We can easily implement  `canvas.get_contex("bitmap")` that mimics [this](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmapRenderingContext). This offers a really simple way for people to e.g. make small games using a numpy array. Simply fun! Also allows creating working examples that don't require tons of lines for wgpu ;)

The contract between the context and the canvas:

* Each canvas backend must support at least *either* the "screen"  or "bitmap"  present methods. It may provide more (specialized) methods.
* When a context is instantiated, it gets a `present_capabilities`, a dict with info on what methods of presentation the canvas supports.
* Each context *must* support *both* "screen"  and "bitmap" present methods. 
  * If a context can render to screen, it's probably a GPU lib, and the texture can be downloaded to cpu.
  * If a context is inherently image-based, it can use a little adapter, that we provide, that presents the image to screen using wgpu.
* The context must implement a method called `present()`, that returns a little dict that specifies what present method was used, and possibly the data corresponding to the present method.
  
Future:

* One use-case for more additional, more specialized, render methods is for remote rendering (e.g. jupyter_rfb). E.g. a present method that sends compressed images or diffs.

almarklein:
Closes #20 

This will break compatibility with wgpu-py. So let's keep this draft for now.

* [x] Rename `_rc_get_present_info()` -> `_rc_get_present_methods()`
* [x] Rename the "image" present method to "bitmap".
* [x] Document what  `_rc_get_present_info()` can/should return.
* [x] Refactor `canvas.get_context()` mechanics. Can now load a context from anywhere. 
* [x] Document the API that a context should have in order to play nice with the above.
* [x] Implement a simple `BitmapRenderingContext`.
* [x] Implement a `BitmapToScreenAdapter` to help contexts like the above to render to screen for canvases that only support the screen present-method. 
* [x] Docs.
* [x] Examples. 
* [ ] Tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ One canvas API, multiple backends 🚀
   <img width=354 src='https://github.com/user-attachments/assets/af8eefe0-4485-4daf-9fbd-36710e44f07c' />
 </div>
 
+*This project is part of [pygfx.org](https://pygfx.org)*
+
 
 ## Introduction
 
@@ -33,7 +35,7 @@ same to the code that renders to them. Yet, the GUI systems are very different
 
 The main use-case is rendering with [wgpu](https://github.com/pygfx/wgpu-py),
 but ``rendercanvas``can be used by anything that can render based on a window-id or
-by producing rgba images.
+by producing bitmap images.
 
 
 ## Installation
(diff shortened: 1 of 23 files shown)


almarklein:
This docs page is I think a good summary of what I'm trying to achieve here: https://rendercanvas--21.org.readthedocs.build/21/contextapi.html

PR #23, opened 2024-11-18
Run examples on ci

almarklein:
No screenshot comparisons yet. Just a little crude version. I think in the long run we need a more streamlines solution for running screenshots / building the gallery, that we can use in rendercanvas, wgpu-py, pygfx, and downstream libs as well.


almarklein:
Oh right, wgpu is not compatible yet (I was on the branc that is, locally).

PR #25, opened 2024-11-18
Handle ctrl-c

almarklein:
See https://github.com/pygfx/pygfx/issues/377

Inspired mostly by https://github.com/encode/uvicorn/blob/master/uvicorn/server.py

Sending sigint will stop the event loop, if we're the one running it. These changes improve the behavior of the base loop implementation, so that the subclasses can actually be made quite a bit simpler.

* [x] Tested on MacOS 
* [x] Tested on Windows
* [x] Tested on Linux


almarklein:
Interrupt (ctrl-c) simply calls `loop.close()`. This will first (attempt to) close all canvases. When all canvases are closed, the loop ends. If for some reason the windows do not close, one can close/interrupt called again to close the loop without closing the windows.

Issue #55, opened 2025-02-04
bug(ish): calling `winId()` may break many other Qt widgets


tlambert03:
it looks like the call to `winId()` here in rendercanvas: https://github.com/pygfx/rendercanvas/blob/5d7385217e4b41ea6313c386c1bca09e91e39107/rendercanvas/qt.py#L255-L275

causes problems with the dock widget system, such that when you drag them, they don't "go with you" ...  It's hard to explain so here's a visual:

before adding pygfx canvas, when you click and drag a dock widget it goes with you:

https://github.com/user-attachments/assets/c7aa8d04-7a0d-496e-aa0e-60615682a604

after adding pygfx canvas, it does indeed float... but then it "stays put" until you go back and click on it again

https://github.com/user-attachments/assets/156f2bb4-cd97-46ff-9753-dcccacee9f01

really not sure who should be "responsible" for this one (pygfx or the end user).  But so far, I haven't yet been able to find a way to recover normal dock widget activity after having instantiated a QtRenderCanvas.

almarklein:
Thanks for the report! I can indeed reproduce this with this minimal example. It does not even need a rendercanvas; creating a widget and calling `winId()` on it is enough to trigger the issue:

```py
from PySide6 import QtWidgets, QtCore
# from rendercanvas.qt import QRenderWidget
# from rendercanvas.utils.cube import setup_drawing_sync


class Main(QtWidgets.QMainWindow):
    def __init__(self):
        super().__init__()
        self.resize(800, 600)

        # self.canvas = QRenderWidget( update_mode="continuous")
        # draw_frame = setup_drawing_sync(self.canvas)
        # self.canvas.request_draw(draw_frame)

        self.canvas = QtWidgets.QWidget()
        self.canvas.winId()

        self.dock1 = QtWidgets.QDockWidget("Dock 1", self)
        self.dock2 = QtWidgets.QDockWidget("Dock2", self)
        self.dock3 = QtWidgets.QDockWidget("Dock3", self)

        for dock in (self.dock1, self.dock2, self.dock3):
            dock.setFeatures(QtWidgets.QDockWidget.DockWidgetFeature.DockWidgetMovable)
            self.addDockWidget(QtCore.Qt.DockWidgetArea.LeftDockWidgetArea, dock)

        self.dock1.setWidget(QtWidgets.QPushButton("A button"))
        self.dock2.setWidget(QtWidgets.QPushButton("Another button"))
[...]

PR #60, opened 2025-03-12
Dont get winId in qt and wx with bitmap present-method

almarklein:
Ref #55

Even calling `winId()` causes issues for Qt widgets. So let's not do it unless we need that id.


almarklein:
An unexpected result: this actually improves the performance significantly! Moving from 30 to 35 FPS on a 4k canvas 🚀

Korijn (approved):

PR #61, opened 2025-03-12
Detect qt winid changes

almarklein:
Adds a bit of logic so changes in winid are detected and reported as a warning. It does *not* act in any other way, because the system currently assumes the winid does not change. Making it update the surface when the winid changes would take quite a bit of refactoring. Before doing that I want to know whether its needed. If a winid change causes issues, people who encounter such issues will see the logger warning.


hmaarrfk:
We are happy to test on Linux when you make a release for this.

PR #65, opened 2025-03-14
Initial size of glfw windows

kushalkolar:
Not sure if this is a good fix but made the PR to get thoughts! The initial canvas size in Qt and jupyter are the same as what is given by the `size` kwarg, but the `glfw` canvas does not use this for the initial size (it appears to resize aftewards?). This is an issue for [creating layouts that are based on the inital canvas size](https://github.com/fastplotlib/fastplotlib/pull/740#issuecomment-2721872300), not sure if this is the best fix.

### DIFF
--- a/rendercanvas/glfw.py
+++ b/rendercanvas/glfw.py
@@ -181,8 +181,13 @@ class GlfwRenderCanvas(BaseRenderCanvas):
         glfw.window_hint(glfw.CLIENT_API, glfw.NO_API)
         glfw.window_hint(glfw.RESIZABLE, True)
 
+        if "size" in kwargs.keys():
+            w, h = kwargs["size"]
+        else:
+            w, h = 640, 480
+
         # Create the window (the initial size may not be in logical pixels)
-        self._window = glfw.create_window(640, 480, "", None, None)
+        self._window = glfw.create_window(w, h, "", None, None)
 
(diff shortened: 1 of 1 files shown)


almarklein:
It does *try* to set the initial size correctly. When the `camvas.__init__()` is done, the size should be correct. In particular, the `_final_canvas_init()` call at the `__init__` sets the size and title of the window. 

That said, the window *is* indeed initially a different size. Maybe it should initialize minimized/hidden.

How do you observe that the window is initially of a different size? Do you receive events or even a draw?

almarklein:
Could you try adding these two lines instead?
```py
# Add this where the other window hints are set
glfw.window_hint(glfw.VISIBLE, False)  # start hidden

# Add this right after self._final_canvas_init()
glfw.show_window(self._window)

```

almarklein:
@kushalkolar, I think I fixed the issue in #70 (it applies the lines I suggested above).

Korijn:
Do we still want to merge this?

almarklein:
No, because #70 fixes it in better way (using the uniform mechanic to set the initial given size). Closing. @kushalkolar feel free to re-open if #70 does not fix it.

kushalkolar:
can confirm works now, thanks! :partying_face:

PR #68, opened 2025-03-23
Fix AttributeError on present() "fail" case

rafaelbeckel:
When the context returns a failure message, the exception handling logic tries to "get" a field from the message string instead of the result dict, causing the error not to be displayed correctly:

`AttributeError: 'str' object has no attribute 'get'`


Korijn (approved):
Thanks for your contribution! ❤️ 

@almarklein makes sense to me.

almarklein (approved):
Nice catch, thanks!

PR #70, opened 2025-03-24
GLFW window start at correct size and draw during resize

### LINKED ISSUE #69: Feature request: interative resize event
Hey team,

I noticed the "resize" event is fired on mouse release.
This causes the surface to stretch while I'm dragging the window border.

Is there a way to make the event fire every frame while the user drags the border?

Here's a recording comparing RenderCanvas to Winit running the same shader:
https://www.loom.com/share/12f9b9adbe324fcda8ae5dc708c35943

almarklein:
Closes #69, relates to #65

* [x] Create the glfw window in hidden state, then show once properly initialized. This avoids starting at a hard-coded size and then resizing.
* [x] Draw during resizing. No events are processed and the loop is on pause during a user-resize-action. But at least it feels a bit more smooth.

### DIFF
--- a/rendercanvas/glfw.py
+++ b/rendercanvas/glfw.py
@@ -147,7 +147,7 @@ def get_physical_size(window):
 
 
 def enable_glfw():
-    glfw.init()
+    glfw.init()  # this also resets all window hints
     glfw._rc_alive = True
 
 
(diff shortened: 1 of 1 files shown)


almarklein:
FWIW its ok if `_draw_frame_and_present()` is called outside of our scheduling mechanics; happens in Qt too.

Korijn (approved):

PR #80, opened 2025-05-08
Add pointer_enter and pointer_leave events

### LINKED ISSUE #78: Pointer enter and leave events
Replicate https://github.com/vispy/jupyter_rfb/pull/116 for our backends.

almarklein:
Closes #78

### Implemented for
* glfw
* qt
* wx

(for jupyter it already works via jupyter_rfb since https://github.com/vispy/jupyter_rfb/pull/116)

### Behavior for all backends

*I made sure the below behavior is implemented in all backends.*

* Send enter / leave event when the mouse enters / leaves the window.
* The canvas does not emit pointer move events when the pointer has left the canvas, unless a button is pressed (i.e. dragging / mouse tracking).
* When the application loses focus, a leave event is emitted.
* When the application regains focus, an enter event is emitted if the pointer if over the canvas. 
* If the application regained focus by clicking on the canvas, that click does not result in pointer events (down, nor move, nor up).
* When the application does not have focus, it does not emit pointer move events.

Exceptions:
* On Wx on MacOS, there are no pointer move events at all when no button is pressed. This is a known issue.
* In glfw, if the window is initialized with the pointer over the canvas, the initial (very first) enter event is emitted when the mouse is moved.
* In the browser, when the application gains focus with the pointer over the canvas, the enter event is emitted when the mouse is moved.

### DIFF
--- a/rendercanvas/_events.py
+++ b/rendercanvas/_events.py
@@ -22,8 +22,10 @@ class EventType(BaseEnum):
     resize = None  #: The canvas has changed size. Has 'width' and 'height' in logical pixels, 'pixel_ratio'.
     close = None  #: The canvas is closed. No additional fields.
     pointer_down = None  #: The pointing device is pressed down. Has 'x', 'y', 'button', 'butons', 'modifiers', 'ntouches', 'touches'.
-    pointer_up = None  #: The pointing device is released. Same fields as pointer_down.
-    pointer_move = None  #: The  pointing device is moved. Same fields as pointer_down.
+    pointer_up = None  #: The pointing device is released. Same fields as pointer_down. Can occur outside of the canvas.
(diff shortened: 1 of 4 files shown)


Korijn:
> I made sure the below behavior is implemented in all backends.

Including jupyter/browser? Is it consistent?

almarklein:
I checked each backend, for a full-window canvas, and checked some basic browser behavior. I did not check the case when a canvas is embedded in an application (he canvas in a notebook can be considered "a canvas in an application" in this context as well). I should check that as well 🤔

almarklein:
Good to go from my end @Korijn

Korijn (approved):

FlynnOConnell:
@almarklein this will fix the behavior I'm seeing below? 


https://github.com/user-attachments/assets/15d963e4-613c-4e1b-88d3-074d67f59122

PR #85, opened 2025-05-12
Type hints for the public API

almarklein:


Vipitis:
maybe a wild question, but do you think there is some way to get type hints when importing from `.auto`? perhaps it should point to the base class to at least see what the args are?

almarklein:
Good call! Fixed 😎

PR #89, opened 2025-05-13
Add examples for simulations

### LINKED ISSUE #49: Updating in a for or while loop
Finally catching up with all this! Taking a look at the examples and wondering if this would be the right way to update the canvas in a for or while loop. It works but I'm not sure if this is the way it's intended to be used, without an event loop running.

```python
import numpy as np
from rendercanvas.auto import RenderCanvas, loop

canvas = RenderCanvas(update_mode="ondemand")
context = canvas.get_context("bitmap")

for i in range(100):
    w, h = canvas.get_logical_size()
    shape = int(h) // 4, int(w) // 4

    bitmap = np.random.uniform(0, 255, shape).astype(np.uint8)
    context.set_bitmap(bitmap)
    canvas.force_draw()
```

related: https://github.com/pygfx/pygfx/discussions/667

I'm starting to think about how to implement this for `fastplotlib`, might help to have a mode where the canvas is updated only when an event occurs or when any buffer in the scene changes? Anyways that will be a separate issue which I'll think about later.

almarklein:
Closes #49

* Do a simulation synced with the animation-loop, so you see every step.
* Do a simulation that runs at its own pace, seeing some steps.

### DIFF
--- a/examples/simulation1.py
+++ b/examples/simulation1.py
@@ -1,0 +1,62 @@
+"""
+Running a simulation, part 1
+----------------------------
+
+In this example, we run a simulation (Conway's Game of Life),
+where each rendered frame is one step in the simulation.
+This is convenient when you want to see every step of an animation.
+
+We could simply call this simulation_step() inside the animate function
+to get one step on each frame. But in this example we wrapped
+the simulation in a generator function.
+
+Note that this example uses the bitmap context, but this can be used
+with wgpu in the same way.
+"""
+
+import numpy as np
+
+from rendercanvas.auto import RenderCanvas, loop
+from scipy.signal import convolve2d
+
+
+canvas = RenderCanvas()
+canvas.set_update_mode("continuous", max_fps=10)
+
+
+context = canvas.get_context("bitmap")
+
+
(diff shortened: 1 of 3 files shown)


almarklein:
@kushalkolar  @clewis7 let me know if this is clear, and if this covers the use-cases you had in mind. Let's leave this open for a while, so you can try this out in FPL, and we can improve the examples along the way.

Vipitis:
hey, I know this is an older example but I just tried it - not sure if this is still the intended way.

I have a compute pass that does the progressive light simulation and then really only care about reading it out to the canvas at like 30fps.

If I put both calls into the draw function I can easily reach 1200fps with fastest and vsync off. But if I do it as an ask task like example2 here the performance is way worse like maybe 100 steps per second. I put the sleep to like `0.00001` but there seems to be other overhead somewhere. I think the best solution for me would be indirect draw and larger dispatches but there is other synchronization issues etc... so I will just use a sinlge draw function for now it's fast enough -just feels inefficient.

As you can't access the scheduler or title info, I also can't put a condition in my draw function to do the copy to canvas after 30ms has passed or something.

PR #92, opened 2025-05-22
Warn for inproper use of WxRenderWidget

almarklein:


tlambert03:
This is fine, however, there’s shouldn’t be technically anything wrong with creating an unparented widget and then reparenting it before starting the event loop right?

Thats what I had been doing and why this warning sorta misses the point (which is more that rendercanvas doesn’t support that mode of operation… even if you didn’t intend to use it as a tool widget)

Vipitis:
I think the same is also true for `QRenderWidget` I had imported that instead of `RenderCanvas` and I had to kill the terminal to recover.

almarklein:
> there’s shouldn’t be technically anything wrong with creating an unparented widget and then reparenting it before starting the event loop right?

Good point!  I tweaked the code so that it allows a (temporary) unparented widget. Could you try and see if it works with what you were trying?

Issue #93, opened 2025-05-22
Catch auto-repeat keys in pyside6


SpicyRicecaker:
Is it possible to block auto-repeat key events in `pygfx` for a qt window? 

---

## Background

On Linux, in both X11 & Wayland, whenever I press and hold a key, a key release event is fired *on key repeat*. Here is a minimal example of `pyside6` demonstrating this behavior on linux:

```python
from PySide6.QtGui import QKeyEvent
from PySide6.QtWidgets import QApplication, QMainWindow
from PySide6.QtCore import Qt
import sys

class MainWindow(QMainWindow):
    def keyReleaseEvent(self, event):
        print(f"Key released: {event.key()} ({Qt.Key(event.key()).name})")
    def keyPressEvent(self, event: QKeyEvent) -> None:
        print(f"Key down: {event.key()} ({Qt.Key(event.key()).name})")

app = QApplication(sys.argv)
window = MainWindow()
window.setWindowTitle("PySide6 Key Up Example")
window.resize(400, 200)
window.show()
app.exec()
```

output (after pressing and holding, but not physically releasing the 'A' key):

```
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
Key down: 65 (Key_A)
Key released: 65 (Key_A)
```

This makes it really hard to implement a fly camera manually, since it's really hard to tell when the player has let go of the forward button. 

After a lot of gpt and trial and error, I narrowed it down to every single qt window on linux having this problem. But `pyside6` has this nice function called `PySide6.QtGui.QKeyEvent.isAutoRepeat()` that automagically detects this key repeat behavior and filters it out. 

[...]

PR #94, opened 2025-05-23
Do not emit repeated `key_down` and `key_up` events

SpicyRicecaker:
This pull request stops os-level repeated `key_down` and `key_up` events from emitting from the `qt`, `glfw`, and `wx` backends to event handlers.

### Rationale

On linux, when the user presses and holds a key, the operating system sends repeat `key_down` and `key_up` events continuously depending on the input rate, even if the user has not physically lifted the key.

This causes significant friction for if a user wants to detect when a key is held and released.

`jupyter_rfb` already [filters these events](https://github.com/vispy/jupyter_rfb/blob/f84110f2f53eca9f7a99484039bced49a4309ebc/js/lib/widget.js#L331-L345) from ever reaching the server framebuffer from the javascript client. This pull request standardizes that behavior on the `qt`, `glfw`, and `wx` backends.

(Also see #93!)

### DIFF
--- a/rendercanvas/glfw.py
+++ b/rendercanvas/glfw.py
@@ -625,7 +625,9 @@ class GlfwRenderCanvas(BaseRenderCanvas):
             "key": keyname,
             "modifiers": tuple(self._key_modifiers),
         }
-        self.submit_event(ev)
+
+        if not action == glfw.REPEAT:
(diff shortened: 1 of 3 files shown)


almarklein (approved):
Beautiful!

PR #96, opened 2025-06-20
fix PageUp key event

### LINKED ISSUE #95: key event for "PageUp" is "Pageup"
I was mapping keys back into codes and run into this. The reference I used that worked for everything else is https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#navigation_keys

Seems to be correct for wx but wrong for glfw and qt, so I guess it's a bug.

Vipitis:
fixes #95 

since the same error exists in `wgpu-py` too, should I also make a PR for that?

found none else: https://github.com/search?q=org%3Apygfx+%22pageup%22&type=code

### DIFF
--- a/examples/cube_qt_trio.py
+++ b/examples/cube_qt_trio.py
@@ -9,7 +9,6 @@ around, e.g. runnning a glfw canvas with the Qt loop does not work so
 well.
 """
 
-# ruff: noqa: E402
 
 import trio
 from rendercanvas.pyside6 import RenderCanvas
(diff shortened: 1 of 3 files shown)


Korijn (approved):

almarklein:
>  since the same error exists in wgpu-py too, should I also make a PR for that?

You could. But since we're moving (slowly) to rendercanvas, I'd say only do it if it solves a problem for yourself :)

PR #109, opened 2025-09-10
Reach FPS closer to max_fps on Windows

### LINKED ISSUE #107: Continuous update mode doesn't reach `max_fps`
so a while back I noticed this behaviour https://github.com/pygfx/wgpu-py/pull/714#discussion_r2085737068 and it has bothered me ever since.

I wanted to look into this a bit more and finally sat down.

basically setting `update_mode="continuous"` and any `max_fps` seems to always undershoot the target by some margin. I waited a couple seconds and it doesn't seem to grow, mostly stabilizes at some non specific fraction. 30 -> ~22, 60 -> ~34, 90 -> ~64. This is on a 165hz screen and if I use `update_mode="fastest` the framerate easily hits 165.0 or over 1000 with `vsyc=False`. As my CPU is about a decade old, this likely contributes - but more performance is clearly possible.

Image

I browsed the code to understand what is happening and maybe find a solution, the following line seems to be misbehaving:
https://github.com/pygfx/rendercanvas/blob/87bfce1295e543bc46f5b1ff87ac268ffbcaf1ad/rendercanvas/_scheduler.py#L133
putting the blocking `time.sleep(delay)` there instead causes the framerates to be much closer to the target 30 -> 29.5, 60 -> 58.7, 90 -> 84.4 which still isn't perfect but like much better. So perhaps there is some async events happening it's interfering with itself? My knowledge on async is quite limited so I wouldn't know what to look for.

I took timestamps before and after that call and tested different framerates, don't feel like I have the best visualization just yet - so hopefully I think of something tomorrow. Here is my [profiling branch](https://github.com/Vipitis/rendercanvas/tree/profile) (excluding the data) but there is overhead due to this.

One more lead I can offer is that framerates abstract away events by already averaging... so because the delay is based on the previous frame, there might be some oscilation, or the vsync event is somehow impacting the async flow - that is something which I thought I spotted in a trace. Could be the canvas itself tho.

let me know if you have some more things to try for me.

almarklein:
Closes #107

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: [tests, release]
     if: success() && startsWith(github.ref, 'refs/tags/v')
+    environment:
+      name: pypi
+      url: https://pypi.org/p/rendercanvas
+    permissions:
+      contents: write
+      id-token: write
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
@@ -222,5 +228,4 @@ jobs:
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
-        user: __token__
-        password: ${{ secrets.PYPI_PASSWORD }}
+        print-hash: true


almarklein:
> You can have one persistent qobject to perform the dynamic dispatch?

Yeah, we can do that. But at this point I seriously wonder if the added complexity is worth it :)

> I thought Slot was no longer necessary in modern pyside

Looks like there are a few edge-cases where it is.

Issue #38, opened 2024-11-22
Backend: HTMLCanvasElement to run on pyodide / pyscript


almarklein:
Should actually not be too hard. Although this is partly so because I kept this idea in my head at every design decision.

Vipitis:
it seems very possible, I don't have the class yet but pyodide allows you to directly call the js methods https://github.com/pygfx/rendercanvas/compare/main...Vipitis:rendercanvas:browser

will tinker some more this week and maybe open a PR. Because `rendercanvas` installs into pyodide directly from micropip it might be an option to provide something really simple for the auto backend there.

PR #115, opened 2025-09-17
Support for Pyodide and PyScript

Vipitis:
#38 

two evenings of tinkering but I can feel a bit of progress. Rendercanvas looks very web inspired, so I am reading a lot of things between the docs, pyodide docs, pyodide source that sound close but are ever so slightly different. Plus I don't have any webdev experience, it's more like a learning opportunity.

for those that want to give it a try - you can essentially just load the .html as a static page with the python script inserted. But you will need to build wheels and load them locally. Installing from pypi via micropip doesn't include these changes. 

some todos:
* [x] Name the new backend "html", "pyodide", or other?
* [x] support all events
    - [x] resize
    - [x] close
    - [x] mouse move doesn't work yet
    - [x] wheel emits warning -> not anymore
    - [x] (should events block navigation?)
    - [x] listen to custom `message` events from the browser (for example to close?)
* [x] demo page should be able to load different examples (maybe multiple?)
* [x] Implement meta-tests 
* [ ] write tests for CI -> [docs](https://pyodide.org/en/stable/development/building-packages-from-source.html#testing-packages-against-pyodide)
* [x] put interactive canvas in docs? (doc builder also builds the wheel)

### DIFF
--- a/rendercanvas/pyodide.py
+++ b/rendercanvas/pyodide.py
@@ -8,6 +8,7 @@ is not required to set the default sdl2 canvas as the Pyodide docs describe.
 
 __all__ = ["PyodideRenderCanvas", "RenderCanvas", "loop"]
 
+import re
 import sys
 import time
 import ctypes
@@ -27,6 +28,7 @@ from js import (
     window,
     ResizeObserver,
     OffscreenCanvas,
+    navigator,
 )
 
 KEYMAP = {
@@ -59,6 +61,11 @@ def buttons_mask_to_tuple(mask) -> tuple[int, ...]:
         if v == "1":
             res += (MOUSE_BUTTON_MAP.get(i, i),)
     return res
+
+
+looks_like_mobile = bool(
+    re.search(r"mobi|android|iphone|ipad|ipod|tablet", str(navigator.userAgent).lower())
+)
 
 
 # The canvas group manages canvases of the type we define below. In general we don't have to implement anything here.
@@ -209,7 +216,8 @@ class PyodideRenderCanvas(BaseRenderCanvas):
             # although they will end up in the 'buttons'. The lost/release will only get fired when all buttons
             # are released/lost. Which is why we look up the original button in our `pointers` list.
             nonlocal last_buttons
-            focus_element.focus({"preventScroll": True, "focusVisible": False})
+            if not looks_like_mobile:
+                focus_element.focus({"preventScroll": True, "focusVisible": False})
             el.setPointerCapture(ev.pointerId)
             button = MOUSE_BUTTON_MAP.get(ev.button, ev.button)
             pointers[ev.pointerId] = (button,)


examples/html_canvas.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def set_bitmap(self, bitmap):
+        # TODO: improve performance https://pyodide.org/en/stable/usage/type-conversions.html#buffers
+        # TODO: avoid memory leak!!!
+        # doesn't really exist? as it's part of the context? maybe we move it into the draw function...
+        self._last_bitmap = bitmap # keep track?
+        h, w, _ = bitmap.shape
+        flat_bitmap = bitmap.flatten()
+        js_array = Uint8ClampedArray.new(flat_bitmap.tolist())
+        image_data = ImageData.new(js_array, w, h)
```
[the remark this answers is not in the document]
Vipitis: pyodide has `.assign` and looks like it works much faster already. Not sure if the other two intermediates could be skipped.

Vipitis:
feel like some stuff breaks due to the python name mangling when using double underscores in combination with the subclass... which happens quite a bit - so I am sorta surprised it still works this far.

But I can't find any such references for Pyodide which would be odd if that is a known limitation.

Vipitis:
> feel like some stuff breaks due to the python name mangling

that wasn't the case. I actually fell through the line here https://github.com/pygfx/rendercanvas/blob/fa7defce2ea84787a5f10f31f9b6bbf08e5a9379/rendercanvas/base.py#L451C17-L451C28 because I hadn't implemented context as a class.

I took a lot of inspiration form the bitmap scripts and got it working. I also registered the auto backend successfully - meaning if you build the wheel and then load it statically the examples `noise.py` and `snake.py` work out of the box (although not events yet). But the weekend has a few more days :)


https://github.com/user-attachments/assets/f2fe5af1-1e5b-43c1-8c52-13c7ba03db0e

Vipitis:
singular keydown event works... so more events and other types shouldn't be impossible. However I will get to that another day.

https://github.com/user-attachments/assets/0ee911db-3043-4822-98fb-1c9bbe1ce9be

Vipitis:
tried all day to make it work for the docs ... but either the .whl don't get included as static files or pyodide has trouble importing the wheel. I also wanted to automate the iframe inclusion with sphinx-gallery but seems like you need to either modify the `gen_rst` function or write an image scaper to append some custom rst to the examples... so I just added the comment to the relevant examples instead 🤷 

classic "works on my machine", so have a video of what could have been instead:

https://github.com/user-attachments/assets/09a444ff-0b67-4de2-8d83-89be10c14634

Korijn:
Even so it's awesome how far you managed to take this!

rendercanvas/html.py
```diff
@@ -0,0 +1,370 @@
...
+
+import sys
+if "pyodide" not in sys.modules:
+    raise ImportError("This module is only for use with Pyodide in the browser.")
+
+# packages available inside pyodide
+from pyodide.ffi import run_sync, create_proxy
+from js import document, ImageData, Uint8ClampedArray, window
+
+# TODO event loop for js? https://rendercanvas.readthedocs.io/stable/backendapi.html#rendercanvas.stub.StubLoop
```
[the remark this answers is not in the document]
almarklein: > so should I drop my loop class and just import the asyncio loop instead?

Yes, let's just do that. It's the simplest to do, and makes sure all works together if the user uses asyncio for other stuff.

docs/backends.rst
```diff
@@ -264,6 +269,47 @@ subclass implementing a remote frame-buffer. There are also some `wgpu examples
...
+    </head>
+    <body>
+    ...
+    <canvas id="canvas" width="640" height="480"></canvas>
+    <script type="text/javascript">
+        async function main(){
+            pythonCode = `
+                # Use python script as normally
+                from rendercanvas.auto import RenderCanvas, loop
+                canvas = RenderCanvas(title="Example")
```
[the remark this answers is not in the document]
almarklein: We have `multiple.py`. And the `qt_app.py` example can be changed relatively easily to embed to canvases.

rendercanvas/html.py
```diff
@@ -84,7 +84,7 @@ def buttons_mask_to_tuple(mask) -> tuple[int, ...]:
                     res += (mouse_button_map.get(i, i),)
             return res
 
-
+        self._pointer_inside = False # keep track for the pointer_move event
```
Korijn, line 87: You might want to look into:

* Pointer capturing
* Mouse leave and enter events

Maybe you already have and this comment is entirely moot

rendercanvas/html.py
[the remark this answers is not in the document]
Vipitis: one more consideration that might be complex to implement: use the browsers `requestAnimationFrame` with a callback to run the loop? I spotted something like this here: https://github.com/pyglet/pyglet/blob/5d1fe49db83bd2c060d97226e1334f06a706b1fc/pyglet/app/async_app.py#L64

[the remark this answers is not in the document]
Vipitis: yeah, with some of the wgpu examples I am having issues where the animation stops randomly or when you change tab etc... it just suspends and doesn't recover. So we might need a webloop or whole sheduler

rendercanvas/html.py
```diff
@@ -0,0 +1,448 @@
...
+                "key": proxy_args.key,
+                "time_stamp": proxy_args.timeStamp,
+            }
+            self.submit_event(event)
+
+        self._key_up_proxy = create_proxy(_html_key_up)
+        document.addEventListener("keyup", self._key_up_proxy)
+
+        # char ... it's not this
+        # def _html_char(proxy_args):
```
almarklein, line 305: Have a look at the implementation of char events in jupyter_rfb https://github.com/vispy/jupyter_rfb/pull/119

docs/backends.rst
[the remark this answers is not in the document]
almarklein: I'm pretty sure we can hook things up in a clean way, but I will need to sit down and play a bit to get it right.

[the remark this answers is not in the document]
Vipitis: feel free to commit into this branch or similar if you have the capacity. I am very much out of my depth with the web stuff and also don't have too much time currently (settling into new job and new university) to sit down for some long evenings and figure it out.

rendercanvas/html.py
Vipitis: I looked at that initially, I tried to avoid actually changing the webpage, but maybe there is no way around it. Which also opens new questions like should the python code spawn new canvases? 
Also all the global events might mess up when there are multiple canvases...

rendercanvas/html.py
```diff
@@ -415,39 +425,30 @@ def _rc_present_bitmap(self, **kwargs):
...
-            loop.call_soon(self._rc_force_draw)
+        # Now present the image data.
+        # For this we can blit the image into the canvas (i.e. no scaling). We can only use this is the image size matches
+        # the canvas size (in physical pixels). Otherwise we have to scale the image. For that we can use an ImageBitmap and
+        # draw that with CanvasRenderingContext2D.drawImage() or ImageBitmapRenderingContext.transferFromImageBitmap(),
+        # but creating an ImageBitmap is async, which complicates things. So we use an offscreen canvas as an in-between step.
+        cw, ch = self._canvas_element.width, self._canvas_element.height
+        if w == cw and h == ch:
+            # Quick blit
+            self._canvas_element.getContext("2d").putImageData(image_data, 0, 0)
```
Korijn: out of curiosity: it is strictly necessary to download the image data to a JS buffer and then push it to a canvas? can't we render to the canvas context directly?

I imagine that's where you get the webgpu context from anyway?

almarklein: This is primarily meant for the "bitmap" context. We could also use it for the wgpu contetx, but the plan is to allow writing to the wgpu context directly.

Vipitis:
Hey @almarklein thanks a lot for finishing this. I closely followed your commits over the last days but will look in detail when I am home again. Haven't looked into PyScript yet, it sounds really useful. 

The docs examples are great to have (I am seeing a lot of potential for FPL, too), however it might be worth considering to not load pyodide and the wheel automatically - could make the docs a little heavy. 

It even worked on my phone - but the Textinput field opens the keyboard which is a bit unexpected. But that's not really the target audience anyway. 

If I find the time I will adapt the wgpu PR, but feel free to commit into that branch too -  if you have the capacity.

docs/backends.rst
```diff
@@ -265,6 +271,78 @@ subclass implementing a remote frame-buffer. There are also some `wgpu examples
...
+
+                    @canvas.request_draw
+                    def animate():
+                        context.set_bitmap(data)
+                `
+            // load Pyodide and install Rendercanvas
+            let pyodide = await loadPyodide();
+            await pyodide.loadPackage("micropip");
+            const micropip = pyodide.pyimport("micropip");
+            await micropip.install("rendercanvas");
```
Vipitis: this example also needs numpy

examples/serve_browser_examples.py
```diff
@@ -0,0 +1,191 @@
...
+                    data = f.read()
+                self.respond(200, data, "application/octet-stream")
+            else:
+                self.respond(404, "wheel not found")
+        elif self.path.endswith(".html"):
+            name = self.path.strip("/")
+            if name in py_examples:
+                pyname = name.replace(".html", ".py")
+                html = pyscript_template.replace("example.py", pyname)
+                html = html.replace('"rendercanvas"', f'"./{wheel_name}"')
```
Vipitis: found a problem where when trying the example from docs... this also replaces the canvas id

image

docs/conf.py
```diff
@@ -64,8 +66,88 @@
...
+print("Copy wheel to static dir")
+shutil.copy(
+    wheel_filename,
+    os.path.join(ROOT_DIR, "docs", "static", wheel_name),
+)
+
+
 # -- Sphinx Gallery -----------------------------------------------------
 
+iframe_placeholde_rst = """
```
Vipitis:
```suggestion
iframe_placeholder_rst = """
```

examples/serve_browser_examples.py
```diff
@@ -0,0 +1,191 @@
...
+import flit
+import rendercanvas
+
+
+# Examples to load as PyScript application
+py_examples = [
+    "drag.html",
+    "noise.html",
+    "events.html",
+]
```
Vipitis: could include `snake.py`

examples/serve_browser_examples.py
```diff
@@ -0,0 +1,191 @@
...
+<html>
+<head>
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <title>example.py via PyScript</title>
+    <script type="module" src="https://pyscript.net/releases/2025.10.3/core.js"></script>
+</head>
+
+<body>
+    <a href="/">Back to list</a><br><br>
+
```
Vipitis: perhaps adding the docstring from the top of the example just as text here?

rendercanvas/pyodide.py
```diff
@@ -0,0 +1,541 @@
+"""
+Support to run rendercanvas in a webbrowser via Pyodide.
+
+User code must provide a canvas that is in the dom, by passing the canvas
+element or its id. By default it searcges an element with id "rendercanvas". It
```
Vipitis:
```suggestion
element or its id. By default it selects an element with id "rendercanvas". It
```

rendercanvas/pyodide.py
```diff
@@ -0,0 +1,541 @@
...
+        self._final_canvas_init()
+
+    def _setup_events(self):
+        # Idea: Implement this event logic in JavaScript, so we can re-use it across all backends that render in the browser.
+
+        el = self._canvas_element
+        el.tabIndex = -1
+
+        # Obtain container to put our hidden focus element.
+        # Putting the focus_element as a child of the canvas prevents chrome van emitting input events.
```
Vipitis:
```suggestion
        # Putting the focus_element as a child of the canvas prevents chrome from emitting input events.
```

rendercanvas/pyodide.py
```diff
@@ -0,0 +1,541 @@
...
+                "ntouches": 0,  # TODO later: maybe via https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent
+                "touches": {},
+                "time_stamp": time.time(),
+            }
+            if not ev.altKey:
+                ev.preventDefault()
+            self.submit_event(event)
+
+        def _js_pointer_lost(ev):
+            # This happens on pointer-up or pointer-cancel. We threat them the same.
```
Vipitis:
```suggestion
            # This happens on pointer-up or pointer-cancel. We treat them the same.
```

rendercanvas/pyodide.py
```diff
@@ -0,0 +1,541 @@
...
+
+        # Note: there is no concept of an element being 'closed' in the DOM.
+
+        def _js_pointer_down(ev):
+            # When points is down, set focus to the focus-element, and capture the pointing device.
+            # Because we capture the event, there will be no other events when buttons are pressed down,
+            # although they will end up in the 'buttons'. The lost/release will only get fired when all buttons
+            # are released/lost. Which is why we look up the original button in our `pointers` list.
+            nonlocal last_buttons
+            focus_element.focus({"preventScroll": True, "focusVisble": False})
```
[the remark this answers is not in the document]
almarklein: oh, nice catch! That typo was copied from jupyter_rfb :)

examples/serve_browser_examples.py
almarklein: Good idea!

examples/serve_browser_examples.py
almarklein: Grrr, yeah I ran into this myself too. Twice. I made the default id 'canvas' again. That fixes the problem and is actually a bit nicer in the end.

Vipitis:
one more thing, I am getting ads on the docs page. But not in the wgpu-py or pygfx docs. Not sure what plan you have and how many projects that covers but perhaps it's just a checkbox somewhere. Just a heads up, I don't really mind myself.

almarklein:
> It even worked on my phone - but the Textinput field opens the keyboard which is a bit unexpected.

Also fixed that. For the drag example the squares are huge, but I'll leave that for another day. 

> one more thing, I am getting ads on the docs page

I had a plan for just 2 projects. I now upped it to three :)

almarklein:
> turning off vsync, however that might not make sense in a browser context anyway.

No, vsync does not apply; we're not rendering to the screen, but to a texture that's used by the browser's compositor to combine it with other content.


> Something that happens to me quite a bit during development was errors during the draw call, which then spam the console

We already shorten messages if they are the same. But we can also hide them, and e.g. only emit every second or so.

PR #120, opened 2025-10-25
Refactor resizing logic and avoid mismatch between last resize event and draw

### LINKED ISSUE #56: Qt render canvas resizes before viewport rect is updated
On Qt it seems like the resize event is sent only after the canvas has already resized, with `pygfx` this causes the renderer to try and draw the scene to a canvas that is smaller than expected. This does not happen on glfw or jupyter_rfb.

For example if we use this example and resize the window to make it smaller we get a `GPUValidationError`:

```python
import imageio.v3 as iio
import PyQt6
from rendercanvas.qt import RenderCanvas, loop
import pygfx as gfx


canvas = RenderCanvas()
renderer = gfx.renderers.WgpuRenderer(canvas)
viewport = gfx.Viewport(renderer)
scene = gfx.Scene()

im = iio.imread("imageio:astronaut.png")[:, :, 1]

image = gfx.Image(
    gfx.Geometry(grid=gfx.Texture(im, dim=2)),
    gfx.ImageBasicMaterial(clim=(0, 255)),
)
scene.add(image)

camera = gfx.OrthographicCamera(512, 512)
camera.local.position = (256, 256, 0)
camera.local.scale_y = -1


def update_rect(*ev):
    viewport.rect = 0, 0, *canvas.get_logical_size()


if __name__ == "__main__":
    viewport.rect = 0, 0, *canvas.get_logical_size()
    canvas.add_event_handler(update_rect, "resize")

    canvas.request_draw(lambda: viewport.render(scene, camera, flush=True))
    loop.run()
```

```python
Draw error
Traceback (most recent call last):
  File "/home/kushal/repos/rendercanvas/rendercanvas/_coreutils.py", line 41, in log_exception
    yield
  File "/home/kushal/repos/rendercanvas/rendercanvas/base.py", line 409, in _draw_frame_and_present
    self._draw_frame()
  File "/home/kushal/repos/pygfx/examples/feature_demo/geometry_image.py", line 43, in <lambda>
    canvas.request_draw(lambda: viewport.render(scene, camera, flush=True))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kushal/repos/pygfx/pygfx/utils/viewport.py", line 78, in render
    self.renderer.render(scene, camera, rect=self.rect, flush=flush)
  File "/home/kushal/repos/pygfx/pygfx/renderers/wgpu/engine/renderer.py", line 592, in render
    command_buffers += self._render_recording(
                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kushal/repos/pygfx/pygfx/renderers/wgpu/engine/renderer.py", line 722, in _render_recording
    render_pass.end()
  File "/home/kushal/repos/wgpu-py/wgpu/backends/wgpu_native/_api.py", line 3289, in end
    libf.wgpuRenderPassEncoderEnd(self._internal)
  File "/home/kushal/repos/wgpu-py/wgpu/backends/wgpu_native/_helpers.py", line 394, in proxy_func
    raise wgpu_error  # the frame above is more interesting ↑↑
    ^^^^^^^^^^^^^^^^
wgpu._classes.GPUValidationError: Validation Error

Caused by:
  In wgpuRenderPassEncoderEnd
    In a set_viewport command
      Viewport has invalid rect Rect { x: 0.0, y: 0.0, w: 1280.0, h: 960.0 }; origin and/or size is less than or equal to 0, and/or is not contained in the render target (1278, 960, 1)
```

almarklein:
Fixes #56

Instead of backends needing to implement `_rc_get_physical_size`, `_rc_get_logical_size` and `_rc_get_pixel_ratio`, they must now call `_set_size_info()` whenever their size changes. Backends should also not submit a resize event, the base canvas now handles that.

The base class will now emit a resize event right before processing other events, but *also* right before doing the draw, if the size has changed since events were processed. This fixes #56.

Another benefit is that the backends become simpler, and the getters like `get_physical_size()` may be faster because values are read from the base class's cache, instead of calculated from the backend.

### DIFF
--- a/rendercanvas/_scheduler.py
+++ b/rendercanvas/_scheduler.py
@@ -169,7 +169,13 @@ class Scheduler:
             if (canvas := self.get_canvas()) is None:
                 break
 
-            # Process events now
+            # Process events now.
+            # Note that we don't want to emit events *during* the draw, because event
+            # callbacks do stuff, and that stuff may include changing the canvas size,
+            # or affect layout in a UI application, all which are not recommended during
+            # the main draw-event (a.k.a. animation frame), and may even lead to errors.
+            # The one exception is resize events, which we do emit during a draw, if the
+            # size has changed since the last time that events were processed.
             canvas._process_events()
 
             if not do_draw:
--- a/rendercanvas/base.py
+++ b/rendercanvas/base.py
@@ -468,6 +468,11 @@ class BaseRenderCanvas:
             if self._rc_get_closed():
                 return
 
+            # Note: could check whether the known physical size is > 0.
+            # But we also consider it the responsiblity of the backend to not
+            # draw if the size is zero. GUI toolkits like Qt do this correctly.
+            # I might get back on this once we also draw outside of the draw-event ...
+
             # Make sure that the user-code is up-to-date with the current size before it draws.
             self.__maybe_emit_resize_event()
 
--- a/rendercanvas/qt.py
+++ b/rendercanvas/qt.py
@@ -564,12 +564,11 @@ class QRenderWidget(BaseRenderCanvas, QtWidgets.QWidget):
         # * On Win10 + PyQt6 the ratio is correct (non-integer).
         ratio = self.devicePixelRatioF()
 
-        # When the ratio is not integer (qt6), we need to somehow round
-        # it. It turns out that we need to round it, but also add a
-        # small offset. Tested on Win10 with several different OS
-        # scales. Would be nice if we could ask Qt for the exact
-        # physical size! Not an issue on qt5, because ratio is always
-        # integer then.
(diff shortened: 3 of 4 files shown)


FlynnOConnell:
🙇‍♂️ thank you!

FlynnOConnell:
Confirmed no more resizing errors! Unless you reduce the size to nothing, which is normally not an issue 


https://github.com/user-attachments/assets/187d16ab-8025-4ec3-8e87-21c2f04797fe

almarklein:
> Unless you reduce the size to nothing

In this case it's FPL that does not apply bounds to the viewport :)

But a zero size to the canvas can in theory also happen. I added a note regarding that.

Issue #114, opened 2025-09-16
No (supported) API to change the pixel_ratio


hmaarrfk:
It would be good to be able to change the pixel ratio dynamically.

I feel like browsers do this with Ctrl+"=" and Ctrl+"-".

AI would recommend doing it as part of `set_logical_size` but perhaps another API is more useful.

Korijn:
> I feel like browsers do this with Ctrl+"=" and Ctrl+"-".

Can you confirm this? I don't think this is true.

hmaarrfk:
I’ll try to share something with you both. 

But as you “zoom” you want to maintain your same number of physical pixels to draw sharply right?

So you must change logical and pixel ratio correct?

almarklein:
Oh wait, I realize there's different use-cases, and I'm not sure which one you mean.

1. Have a canvas using a glfw / qt window, and being able to "zoom", similar as in a browser, so the font gets larger, lines thicker etc. -> Yes, we could add that, I can see it being useful.

2. Having an offscreen canvas that you want to match some remote canvas. As the remote's pixel ratio changes, you want the offscreen canvas to do too. Right now, you can set the pixel ratio of the offscreen canvas when you create it, but not change it at runtime. I think it would be quite trivial to add this. Other canvases can change their pixel-ratio too, e..g as you move them from one monitor to another.

The 2nd case is quite easy to implement. The first is also worth implementing, but requires a bit more work.

hmaarrfk:
I feel like doing 1 requires 2 to be implemented. 

Then you bind a shortcut to it. 

I’m ok with pyGFX not having shortcuts. 

Tbh. Setting the `_pixel_width` directly seems to have few negative consequences. 

I am primarily thinking of use case 2

almarklein:
I called it "zoom", because that's what my browser calls it:

<img width="351" height="369" alt="Image" src="https://github.com/user-attachments/assets/bee05a95-85c5-43d2-be6a-f1c24c4838e3" />

But within our context that term may indeed be a bit generic ...


> Logical pixels seem to be integers.

Logical pixels are float.

PR #122, opened 2025-10-29
Added support for set_pixel_ratio to offscreen canvas

almarklein:
Ref #114

cc @hmaarrfk


almarklein:
@hmaarrfk I was thinking the same thing this morning; we also want a way to explicitly set the physical size. I will give this a second round.

BTW: the logical size can in fact be fractional.

almarklein:
Updated.

PR #127, opened 2025-11-03
Implement own context classes

almarklein:
Implement the context classes in `rendercanvas` itself, and *wrap* `wgpu.GPUCanvasContext` instead.

In principle we are free to implement whatever API we want here. We could even merge the context API with the canvas object to obtain a simpler API. But in the end, the concept of a context object that provides a certain way to send the rendered result into the canvas, is actually quite nice. And scalable to other methods. Plus it is familiar to people who know JS. 

Interestingly, the contexts here can be be much closer to the WebGPU API than that of `wgpu-py`, because in this case we do have a 'fixed concept' of what a canvas is, and we automate the `set_physical_size`and `present` methods.

* [x] Implementation
* [x] Docs
* [x] Add test to make sure that wgpu is not needed if not used.
* [x] Naming of present-method and context-type: 'wgpu' vs 'screen'.
* [x] Manually test against https://github.com/pygfx/wgpu-py/pull/764
  * [x] A method to close the context?
  * [x] A function to get the CanvasContext class for the current backend.
  * [x] We need to pass vsync via the `present_info` now.

### DIFF
--- a/rendercanvas/_version.py
+++ b/rendercanvas/_version.py
@@ -50,28 +50,34 @@ def get_extended_version() -> str:
     # Sample first 3 parts of __version__
     base_release = ".".join(__version__.split(".")[:3])
 
-    # Check release
-    if not release:
-        release = base_release
-    elif release != base_release:
-        warning(
-            f"{project_name} version from git ({release})"
-            f" and __version__ ({base_release}) don't match."
-        )
+    # Start version string (__version__ string is leading)
+    version = base_release
+    tag_prefix = "#"
 
-    # Build the total version
-    version = release
+    if release and release != base_release:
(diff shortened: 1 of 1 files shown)


almarklein:
@Korijn this is about ready in terms of API etc.

I gave the context properties to get size info.

I also added `.get_wgpu_context()`, because its type annotation can be `WgpuContext`, which makes for a better experience in tools that do static typing.

I still have to test it against your branch, and e.g. let wgpu-py expose a public function to get the CanvasContext class for the appropriate backend. Will do that tomorrow or so.

almarklein:
I locally confirmed that it works with https://github.com/pygfx/wgpu-py/pull/764.

PR #136, opened 2025-11-10
bump pyscript in docs and examples

Vipitis:
When using the release from pypi, pyscript showed an error like this 
image

I looked around and it appears like they have a list of "official" pyodide packages to check against. I am not sure how to be included there, maybe https://pyodide.org/en/stable/development/adding-packages-into-pyodide-distribution.html#adding-packages-into-pyodide-distribution. Although our implementation is pure python and doesn't need a recipe, so the pypi wheels are fine as they are.

Luckily there was a pyscript update just hours ago that addressed exactly this problem. Now shows a warning but works as expected. 

image

This doesn't impact the docs or the serving script as that points to a local wheel, but I ran into this in the wgpu browser branch.

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -292,7 +292,7 @@ An example using PyScript (which uses Pyodide):
     <html>
     <head>
         <meta name="viewport" content="width=device-width,initial-scale=1.0">
-        <script type="module" src="https://pyscript.net/releases/2025.10.3/core.js"></script>
+        <script type="module" src="https://pyscript.net/releases/2025.11.1/core.js"></script>
     </head>
     <body>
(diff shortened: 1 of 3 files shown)


almarklein (approved):

almarklein:
> it appears like they have a list of "official" pyodide packages to check against.

BTW, was this a problem of Pyodide or of PyScript?

*edit: nevermind, the error message says pyodide*

Vipitis:
Pyodide was fine, I think this was a specific issue with PyScript. Not sure about the wasm wheels. I think there is wheels for it, plus also new ideas on how to distribute python with platform specific libs: https://wheelnext.dev/ or https://docs.pyx.dev/

PR #146, opened 2025-11-17
Add `loop.call_soon_threadsafe()` and re-implement precise sleep

almarklein:
Revisiting #109 (and #107)

## Intro

In #109 I opted for a solution that did some `time.sleep()` in the scheduler. This is a relatively simple solution, that fixed the problem of lower-than-max-fps framerates for all backends.

This morning I was happily planning to implement the polling for wgpu, so that we can properly leverage async, but I bumped in this problem again. I decided to fix the problem in a per-backend manner, so that we can `await sleep()`  with precise timers everywhere (and for the wgpu-polling).

## Changes

* [x] Add `loop.call_soon_threadsafe()`.
  * All loop backends must implement  `_rc_call_soon_threadsafe()`.
* [x] Implement `_coreutils.call_later_from_thread()`
   * This uses a single global thread to handle all call-later needs.
   * The thread is created on first use, because Pyodide cannot do threads.
   * Has about 1ms precision.
* [x] For asyncio and trio, use this util to implement precise sleep on Windows.
* [x] For the wx backend, use this util to implement precise call_later on Windows.
* [x] For the qt backend, use PreciseTimer to implement precise call_later on Windows.
  * Also tried the thread util, but the PreciseTimer seems even more precise.
* [x] The `raw` loop is now dead-simple, because the util takes care of all the scheduling.
* [x] The scheduler now uses a normal `await sleep(..)`, without any trickery.

## Some low-level details

*Mostly to record design decisions. You can probably skip this.*

Some more details about how polling led me here .... I thought about a few different approaches to do the polling:
<details>

* Let wgpu-py run the polling in a thread. But then that thread will call `loop.call_soon()`, which means that the `call_soon()` of all backends will have to be thread-safe, which is currently not the case. 
* I also considered a thread in rendercanvas to call a registered function at a precise interval, which has the same problem.
* So then a nice and clean async task that sleeps regular intervals. This works, but then we run into the issue that async-sleep is inprecise... so here we are.

</details>

Some details about an earlier posted solution that uses a threadpool, see [this link](https://textual.textualize.io/blog/2022/12/30/a-better-asyncio-sleep-for-windows-to-fix-animation/).

<details>
When using a thread pool, one thread is used to sleep for `delay` seconds, and when it wakes up, it will scheduler a callback in the main loop. The downside is that when multiple tasks are running (and sleeping), there will be one thread per task. And when the threads run out, they will wait for each-other and the sleep will be much longer. That's why I came up with a technique that uses a single thread with a priority queue.
</details>

## Some findings while I was experimenting

*Mostly to record interesting findings. You can probably skip this.*

I tried several things to try to find a mechanism that does not suffer from imprecise timer. BTW, ChatGTP is absolutely worthless for this topic.

### Allow a thread to wait for a signal from another thread on a timeout

<details>

This is a simple mechanism that would have been useful. But it looks like its impossible. Using `time.sleep()` is precise, but  any method that has a timeout (`threading.Event.wait()`, `Queue.get()`, `threading.Lock.acquire()`, using `select`  on a socket) is inprecise, and waits 15.6 ms on average. 

```py
import time
import threading
from queue import Queue,SimpleQueue, Empty

import socket
import select


class PollThread(threading.Thread):

    def __init__(self):
        super().__init__()
        self.e = threading.Event()

        self.pair =socket.socketpair()
        self.q = Queue()
        self.lock = threading.Lock()


    def run(self):
        timeout = 0.002

        while True:
            x = []
            for i in range(20):

                t0 = time.perf_counter() 

                time.sleep(timeout)  # This is precise, the rest is not

                # self.e.wait(timeout)

                # try:
                #     self.q.get(True, timeout)
                # except Empty:
                #     pass

                # ready, _, _ = select.select([self.pair[0]], [], [], timeout)

                # self.lock.acquire(True, timeout)

                t1 =  time.perf_counter()
                x.append(t1-t0)

            print(f"{1000*sum(x)/len(x):0.1f}  {1000*max(x):0.1f}")
            x.clear()

poller = PollThread()
poller.start()
```

</details>

### Let a sub-thread invoke a method in the main thread

<details>

This actually *does* work precise!

For Asyncio, you can take a `asyncio.Event`, let a task wait for it, and then let the worker thread do `loop.call_soon_threadsafe(event.set)`.

For Qt it works with a litle code like this:
```py
from PySide6.QtCore import QObject, Signal, Slot
from PySide6.QtGui import QGuiApplication
class InvokeMethod(QObject):
    def __init__(self, method):     
        super().__init__()
        main_thread = QGuiApplication.instance().thread()
        self.moveToThread(main_thread)
        self.setParent(QGuiApplication.instance())
        self.method = method
        self.called.connect(self.execute)
        self.called.emit()

    called = Signal()

    @Slot()
    def execute(self):
        self.method()        
        self.setParent(None)  # trigger garbage collector
```

And then the worker thread simply does `InvokeMethod(self.cb)`.

I considered an approach where `rendercanvas` would do the scheduling in a thread and then use this mechanism to make it do stuff in the main thread. But that would mean things need to be differen on systems where we have no threads, like Pyodide.

</details>

### Piece of code to check that it works

<details>

```py
import time
from rendercanvas.glfw import RenderCanvas
from rendercanvas.asyncio import loop   # use pyside6, asyncio, raw, wx, trio
from rendercanvas.utils.asyncs import sleep


RenderCanvas.select_loop(loop)
c = RenderCanvas()

async def main():
    while True:
        x = []
        for i in range(20):

            t0 = time.perf_counter()
            await sleep(0.002)
            t1 =  time.perf_counter()
            x.append(t1-t0)

        print(f"{1000*sum(x)/len(x):0.1f}  {1000*max(x):0.1f}")
        x.clear()

loop.add_task(main)

loop.run()
```

</details>

### DIFF
--- a/rendercanvas/_coreutils.py
+++ b/rendercanvas/_coreutils.py
@@ -108,9 +108,18 @@ def weakbind(method):
 class CallLaterThread(threading.Thread):
     """An object that can be used to do "call later" from a dedicated thread.
 
-    Care is taken to realize precise timing, so it can be used to implement
-    precise sleeping and call_later on Windows (to overcome Windows' notorious
-    15.6ms ticks).
+    This is helpful to implement a call-later mechanism on some backends, and
+    serves as an alternative timeout mechanism in Windows (to overcome its
+    notorious 15.6ms ticks).
+
+    Windows historically uses ticks that go at 64 ticks per second, i.e. 15.625
+    ms each. Other platforms are "tickless" and (in theory) have microsecond
+    resolution.
+
+    Care is taken to realize precise timing, in the order of 1 ms. Nevertheless,
+    on OS's other than Windows, the native timers are more accurate than this
+    threaded approach. I suspect that this is related to the GIL; two threads
+    cannot run at the same time.
     """
 
     Item = namedtuple("Item", ["time", "index", "callback", "args"])
@@ -140,6 +149,7 @@ class CallLaterThread(threading.Thread):
         wait_until = None
         timestep = 0.001  # for doing small sleeps
         leeway = timestep / 2  # a little offset so waiting exactly right on average
+        leeway += 0.0005  # extra offset to account for GIL etc. (0.5ms seems ok)
 
         while True:
             # == Wait for input
@@ -152,7 +162,7 @@ class CallLaterThread(threading.Thread):
                 # we wait shorter, and then go in a loop with some hard sleeps.
                 # Windows has 15.6 ms resolution ticks. But also on other OSes,
                 # it benefits precision to do the last bit with hard sleeps.
-                offset = 0.016 if is_win else 0.004
+                offset = 0.016 if is_win else timestep
                 try:
                     new_item = q.get(True, max(0, wait_until - perf_counter() - offset))
                 except Empty:
--- a/rendercanvas/qt.py
+++ b/rendercanvas/qt.py
@@ -195,7 +195,7 @@ _show_image_method_warning = (
 
 
 class CallbackWrapperHelper(QtCore.QObject):
-    """Little helper for the high-precision-timer call-laters."""
+    """Little helper for _rc_call_later with PreciseTimer"""
 
     def __init__(self, pool, cb):
         super().__init__()
@@ -211,7 +211,7 @@ class CallbackWrapperHelper(QtCore.QObject):
 
 
 class CallerHelper(QtCore.QObject):
-    """Little helper class for the threaded call-laters."""
+    """Little helper for _rc_call_soon_threadsafe"""
 
     call = Signal(object)
 


almarklein:
cc @Vipitis

almarklein:
I ❤️ this piece of code: 

```py
class RawLoop(BaseLoop):
    ...

    def _rc_run(self):
        while not self._should_stop:
            callback = self._queue.get(True, None)
            try:
                callback()
            except Exception as err:
                logger.error(f"Error in RawLoop callback: {err}")

    def _rc_call_later(self, delay, callback):
        call_later_from_thread(delay, self._rc_call_soon_threadsafe, callback)

    def _rc_call_soon_threadsafe(self, callback):
        self._queue.put(callback)
```

PR #151, opened 2025-11-25
The asyncadapter util now has get_running_loop

almarklein:
*I accidentally merged this, reverted in #155*

The asyncadapter is used by loops that are not async by default, like Qt and Wx. This PR adds a `get_running_loop()`  function to that module, so we can integrate better with sniffio. In particular, in wgpu-py I want to be able to do:

```py
module_name  =  sniffio.current_async_library()
mod = sys.modules[module_name]
loop = mod.get_running_loop()
loop.call_soon_async(...)
```

This code works the same for both asyncio and our adapter. (Trio works differently; we'll handle that with an if statement.)

----

I explored several options for "integrating" the rendercanvas loop with wgpu-py. My first inclination was to go for something like `rendercanvas.get_current_loop()`, but it turns out that it's hard to establish when a loop becomes 'active'. E.g. it may already be running in an interactive session. So it felt like it was hard to keep this properly in sync.

The current solution that goes via asyncio, is nice, because you get exactly the loop for the current code invocation, which makes it very robust, even if ppl go do weird things, like running a loop from another loop.

### DIFF
--- a/tests/test_sniffio.py
+++ b/tests/test_sniffio.py
@@ -11,6 +11,7 @@ than that it's API is very similar to asyncio.
 # ruff: noqa: N803
 
 import sys
+import asyncio
 
 from testutils import run_tests
 import rendercanvas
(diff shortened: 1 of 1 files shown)


almarklein:
Crap, I merged the wrong PR 😅

PR #152, opened 2025-12-04
Improvements to loop lifecycle

almarklein:
## Summary

* [x] The loop is now much more explicit about when it starts and when it stops. (This was hard to get right, with all the subtle differences in backends.)
* [x] Running tasks are cancelled when the loop stops.
* [x] Unfinished async generators are closed when the loop stops, using `set_asyncgen_hooks()`
* [x] Add tests to make sure the state is tracked correctly.
* [x] Add tests that make sure tasks are cancelled.
* [x] Added test to make sure generatores are closed.

## Context

In looking for an alternative approach to sniffio, I ran into `sys.set_asyncgen_hooks()` and `sys.get_asyncgen_hooks()` part of [pep-0525](https://peps.python.org/pep-0525/). From what I understand, our async-adapter should use those to make sure that any unfinished async adapters get cleaned up properly.

So I wrote some tests ... but they already pass, and I have not (yet) been able to come up with a test that fails with the current implementation (which does not use `get_asyncgen_hooks()`).

Nevertheless good to have these tests in place 🤷

### DIFF
--- a/rendercanvas/_loop.py
+++ b/rendercanvas/_loop.py
@@ -27,9 +27,6 @@ HANDLED_SIGNALS = (
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
 
-STATEMAP = {0: "off", 1: "ready", 2: "active", 3: "interactive", 4: "running"}
-STATEMAP_REVERSED = {s: i for i, s in STATEMAP.items()}
-
 
 class BaseLoop:
     """The base class for an event-loop object.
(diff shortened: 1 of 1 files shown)


rendercanvas/_loop.py
```diff
@@ -58,6 +59,7 @@ def __init__(self):
         self.__state = (
             0  # 0: off, 1: ready, 2: detected-active, 3: inter-active, 4: running
         )
+        self._asyncgens = weakref.WeakSet()
```
Korijn: If possible to avoid weak refs of any kind, please do :) especially in hot code paths like event loops, do we really want to have them?

almarklein: My plan is to replicate what asyncio does, and it does use a weakref ;)  but note that this code only runs before a loop start, not on every step.

almarklein:
Current CI failures show that the current implementation suffers from the problem that `set_asyncgen_hooks` were designed to solve. 

I must say that the issue seems pretty niche: it applies to an async generator that is unfinished, and that has a ref somewhere, preventing the gc from cleaning it up. But fair enough.

PR #157, opened 2025-12-12
Drop sniffio and improved limetime

almarklein:
* [x] add `rendercanvas.get_running_loop()` that returns the running loop for the current thread.
* [x] Instead of sniffio use `asyncgen_hooks` to detect the current framework.
* [x] The asyncadapter still registers with sniffio, if sniffio can be imported, to remain compatibility for libs that use sniffio.
* [x] Some more improvements to the loop lifecycle.    

---

I ran some benchmarks
```py
import sys
import time
import threading
import sniffio

n = 1000000


#

t0 = time.perf_counter()
for _ in range(n):
    try:
        sniffio.current_async_library()
    except sniffio.AsyncLibraryNotFoundError:
        pass
t1 = time.perf_counter()
print(' sniffio:', t1 - t0)


#

def detect_current_async_lib():
    ob = sys.get_asyncgen_hooks()[0]
    if ob is not None:
        try:
            libname = ob.__module__.partition(".")[0]
        except AttributeError:
            return None
        if libname == "rendercanvas":
            libname = "rendercanvas.utils.asyncadapter"
        return libname


t0 = time.perf_counter()
for _ in range(n):
    detect_current_async_lib()
t1 = time.perf_counter()
print('asyncgen:', t1 - t0)
```

This shows that using `asyncgen_hooks` is faster. It's speed is also more stable than sniffio, which speed depends on the order it tries different libs.

### DIFF
--- a/rendercanvas/_loop.py
+++ b/rendercanvas/_loop.py
@@ -573,7 +573,7 @@ class BaseLoop:
         raise NotImplementedError()
 
     def _rc_add_task(self, async_func, name):
-        """Add an async task to the running loop.
+        """Add an async task to this loop.
 
         True async loop-backends (like asyncio and trio) should implement this.
         When they do, ``_rc_call_later`` is not used.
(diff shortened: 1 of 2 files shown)


almarklein:
I'm so done with these loops ...

PR #160, opened 2025-12-18
Fix asyncadapter.Event

almarklein:
### The bug

There is a bug in the asyncadapter's `Event` object; `await event.wait()` returns immediately. This event object is only really used in `await sleep()` when `USE_THREADED_TIMER` is set, which is default on Windows.

The consequence is that it breaks the scheduling when on Windows and using the `asyncadapter` (using Qt, Wx, or Raw loop).

I apologize for those affected; this should not have happened. I did run with `USE_THREADED_TIMER`, but apparently only in combination with asyncio. I should have run tests on Windows, or better yet, run Windows-tests on CI, but I was too eager to move forward, because this async stuff was taking sooo long ...

### The solution

* [x] The fix is relatively simple. 
* [x] I also added a Windows and MacOS run to CI to prevent similar issues in the future. 
* [x] Added a unit test to explicitly unit-test the asyncioadapter basics.
* [x] To let the MacOS tests pass, some tests that do timings needed a bit of leeway (github's runners might be busy so stuff runs somewhat slower than locally).

----

Ref https://github.com/pyapp-kit/scenex/pull/42#issuecomment-3667545098

### DIFF
--- a/tests/test_scheduling.py
+++ b/tests/test_scheduling.py
@@ -179,7 +179,7 @@ def test_scheduling_fastest():
     # And after 0.1 s we have a lot more draws. max_fps is ignored
     canvas.draw_count = canvas.events_count = 0
     canvas.active_sleep(0.1)
-    assert canvas.draw_count > 20
+    assert canvas.draw_count > 16  # be gentle for ci
     assert canvas.events_count == canvas.draw_count
 
     # Forcing a draw has direct effect


rendercanvas/utils/asyncadapter.py
```diff
@@ -41,12 +41,12 @@ def __init__(self):
...
         if self._is_set:
-            return
+            pass
         else:
-            return self  # triggers __await__
+            await self  # triggers __await__
 
     def __await__(self):
-        return {"wait_method": "event", "event": self}
+        yield {"wait_method": "event", "event": self}
```
almarklein: The actual fix.

PR #162, opened 2025-12-19
Streamline internal sleep and call_later

almarklein:
Relates to #146

This simplifies the code paths related to async sleep and call_later a bit. In particular:

* Backends use normal timers for call_later. Yes this is imprecise on Windows, but that's expected.
*  The `rendercanvas.asyncs.sleep()` always uses a normal sleep. Yes, imprecise on Windows, but expected.
* Add `rendercanvas.asyncs.precise_sleep()` for where we need precise sleep: one instance in the scheduler.
  * On not-windows, this does the same as `sleep()` 
  * On windows, this uses `call_later_in_thread` in combination with `call_soon_threadsafe`.

Why? I was looking at the code paths related to sleep, the `call_later_in_thread`, and the `_rc_call_later` in the backends. I found the code paths pretty confusion, also which code takes care of having a precise sleep. In fact, the qt precise timer was never actually used, because for that to work, `asyncs.sleep` would have to detect that the qt loop is active and then *not* use the threaded timer but the asyncadapter's sleep. Doing that felt like it would become even more complex. 

This PR also contains a few minor tweaks that I added when investigating yesterdays bug, but forgot to push to the pr.

### DIFF
--- a/tests/test_asyncs.py
+++ b/tests/test_asyncs.py
@@ -22,7 +22,6 @@ loop_classes = [RawLoop, AsyncioLoop, TrioLoop]
 
 @pytest.mark.parametrize("SomeLoop", loop_classes)
 def test_sleep(SomeLoop):
-
     leeway = 0.20 if os.getenv("CI") else 0
 
     times = []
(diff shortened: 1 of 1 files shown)


almarklein:
Just to be clear:
* The scheduling to reach target FPS works just like it did before (now using `precise_sleep` in the scheduler).
* What will be less precise now (on Windows):
   * code that uses `await sleep(..)` from `rendercanvas.utils.asyncs`
   * code that uses `loop.call_later()`
* Does that matter? I think not, because its expected behavior in Windows. For cases where the precision is needed, users can use the new `precise_sleep`.

PR #175, opened 2026-02-02
Qt and wx default to bitmap present

almarklein:
To avoid the many problems users face with presenting to a surface in a qt application, I propose we default to 'bitmap' for qt and wx. I briefly considered detecting cases where problems are more likely to occur, like is it on Wayland, or is the widget in a dockwidget, etc. but I figured we can better keep it simple and consistent.

I feel that with #138 we can now do this. Leaving this open for a bit so ppl can react if they have questions or objections.

Some related issue:
* https://github.com/pygfx/wgpu-py/issues/688
* https://github.com/pygfx/rendercanvas/issues/55
* https://github.com/pygfx/rendercanvas/issues/62

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -192,14 +192,14 @@ But the other way around, running a Qt canvas in e.g. the trio loop, works fine:
 
 
 There are known issue with Qt widgets that render directly to screen (i.e. widgets that obtain ``widget.winId()``),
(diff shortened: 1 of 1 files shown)


almarklein:
OMG, ok, the difference seems to be that Qt applies a color transformation when on a P3 display. I created an issue to track this further https://github.com/pygfx/rendercanvas/issues/176

PR #196, opened 2026-03-13
Resize and title support for Pyodide canvases

almarklein:
Instead of connecting the pyodide canvas to a `<canvas>`, it can now also be connected to a `<div class='renderview-wrapper'>`. When this is done, the a canvas will be created inside it, plus additional elements to support manual resizing, and a titlebar.

image

*You can grab the bottom-right corner and drag it.*

An additional benefit is that content placed inside the wrapper will be deleted when the canvas loads; so its very easy to show a loading notification.

Side note: I started out with just replacing the `<canvas>` with a wrapper element, but that will cause all sorts of problems if users style their canvas, e.g. to position it. We'd have to transfer styles onto the wrapper and that will just become a mess. So I made it explicit. 

If a `<canvas>` is provided (the old approach), the JS code will not change its style at all (except width and height), providing a magic-less option for use-cases that need full control over the canvas.

The Pyodide canvas got a few new methods (the anywidget backend will have these too):
* `set_css_width()` and `set_css_height()`
* `set_resizable()`
* `show_titlebar()`

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -290,9 +290,8 @@ additional dependencies. Currently only presenting a bitmap is supported, as
 shown in the examples :doc:`noise.py <gallery/noise>` and :doc:`snake.py<gallery/snake>`.
 Support for wgpu is underway.
 
-An HTMLCanvasElement is assumed to be present in the
-DOM. By default it connects to the canvas with id "canvas", but a
-different id or element can also be provided using ``RenderCanvas(canvas_element)``.
+The backend will render to an HTML ``<canvas>``. This can be provided with ``RenderCanvas(canvas_element='canvas')``,
+either by providing the element as an object, or via it's id. By default, it connects with the element with id "canvas".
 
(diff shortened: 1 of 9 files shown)


almarklein:
Some technical details on the DOM and CSS, for those interested:

```html
<div class='renderview-wrapper'>
  <canvas />  <!-- or <img>  -->
  <div class='rendervier-top'> 
    <span>The title </span
  </div>
  <div class='renderview-resizer'></div>
</div>
```

The wrapper represents the viewport of the canvas. You can size the wrapper, and the canvas will follow (it has width and height set to 100%). 

The resize element is an invisible 14x14 square absolutely positioned in the bottom-right corner of the wrapper (on top of the canvas). Good to mention that a canvas cannot have child elements.

The titlebar is a bit more tricky; it's absolutely positioned at the top of the wrapper, with a negative offset, so it extends beyond the wrapper. The wrapper sets a `margin-top` to avoid the titlebar from overlapping with other content.

PR #199, opened 2026-03-25
Update renderview.js

almarklein:
The `renderview.js` now has its own home at https://github.com/pygfx/renderview. It defines the event spec (formerly jupyter_rfb event spec) as well as the JS implementation. It also has a tester there, so we can make improvements there, and then copy the source to rendercanvas and jupyter_rfb.

In the new event spec, some fields were renamed. The idea is to propagate these changes into rendercanvas as well, but in small steps, while keeping things backwards compatible.

The new JS code also has some small improvements that were added while working on the jupyter_rfb refactoring.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,5 +133,5 @@ that uses `standardjs`. To format and lint your code locally:
 
 * Install npm if you don't have it already.
 * `npm install --global standard`
-* `standard rendercanvas`  (run this before committing changes to the JavaScript code)
+* `standard rendercanvas --fix`  (run this before committing changes to the JavaScript code)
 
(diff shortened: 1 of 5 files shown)


almarklein:
@Vipitis @Korijn I'm just moving forwards with these JS related PRs to keep up the pace. Feel free to comment also on merged PRs.

Vipitis:
sure, I have been following the progress and even tried the new jupyter_rfb in vscode today - it works with fpl again! (although there might have been some issues with imgui and scrolling, will file issues if I find the time to investigate).

I have no js experience so I won't be able to comment much and I will be away the next two weeks, want to try getting the wgpu pyodide branch out of draft before that,

PR #200, opened 2026-03-25
Add anywidget backend

almarklein:
This adds a backend based on `anywidget`.

It is very similar to the `jupyter` backend (which uses `jupyter_rfb`), but it has no dependencies other than anywidget. This makes it more readily available.

The code is quite similar to `jupyter_rfb`:
* Both use `renderview.js` (and we already have that JS module for pyodide)
* The `renderview-afm.js` is also shared.
* The model uses the same traits, but in the `anywidet` backend they are private.

The main differences:
* A slightly simpler Python implementation, see next points.
* Tighter coupling because the jupyter_rfb abstraction does not sit in between.
* No output context to try to fix the problem of jupyter ignoring sys.stdout and exceptions raised in comm handlers. We don't need it, because our comm/trait handlers feed into the scheduling and only touch user code in fresh asyncio tasks with error handlers.


almarklein:
@kushalkolal do you have an idea to what degree people make use of jupyter_rgb's snapshot utility?
 
If this new backend becomes the default for notebooks, these people can still explicitly import the `jupyter` backend, but I'm not sure how much havoc that will cause :)

almarklein:
Another idea: rename the `jupyter` backend to the `jupyter_rfb` backend, to make it more explicit what the backend does.

kushalkolar:
> @kushalkolal do you have an idea to what degree people make use of jupyter_rgb's snapshot utility?
> 
> If this new backend becomes the default for notebooks, these people can still explicitly import the `jupyter` backend, but I'm not sure how much havoc that will cause :)

In the fastplotlib notebook test we actually [use snapshot](https://github.com/fastplotlib/fastplotlib/blob/4a13244994d2e2ba3ccda11c42f32327354fc7ad/examples/notebooks/nb_test_utils.py#L117), but the notebook test has been broken for a while with the latest `rendercanvas` changes and I will fix them after all this is settled. 

Perhaps the havoc is worth it since it's a major improvement?

FlynnOConnell:
@almarklein FWIW I don't have anyone using rfb snapshots and I don't use them myself

almarklein:
For the record, as part of all this work, I also refactored `jupyter_rfb` to use `anywidget` and `renderview.js`. But jupyter_rfb has become much 'thinner': no build step, most JS delegated. So it makes sense (I think) to just implement it here directly (and maybe also in Vispy).

The advantages are mainly:

- No extra dependencies.
  - if someone installs rendercanvas, and already uses a notebook, chances are big they have anywidget installed, so things Just Work.
  - Not sure how big this advantage is in practice, but it could simplify downstream docs and pyproject files.
- This implementation is a bit simpler than `jupyter_rfb`.
- We want to experiment with JPEG/MPEG encoding on the GPU, so we'd need specialized JS for the decoding. It's much easier to add that to our own backend than to jupyter_rfb, also since this feature would be WGPU specific.

One issue is that there are now two backends for notebooks, which may cause some confusion. I'd propose to make this the new default and keep the jupyter_rfb backend for a while in case people specifically want to use it.

Regarding snapshots, I've thought about it some more, and I think I want to keep them (i.e. port them to this PR), but in a somewhat different form than `jupyter_rfb`  currently does. See https://github.com/vispy/jupyter_rfb/issues/146

FlynnOConnell:
> See [vispy/jupyter_rfb#146](https://github.com/vispy/jupyter_rfb/issues/146)

As you discuss here, the snapshot has been a useful placeholder saying "there are outputs here, you just can't view them in this context". I have been taking this feature for granted ... 

> In this way, ppl can use the canvas interactively, but also easily build tutorials with it that can be statically viewed.

If this would allow the snapshot to include a small interactive video, that would be immensely useful for making tutorials where I can just render videos directly from the notebook outputs rather than saving a video and rendering it with sphinx.

almarklein:
> f this would allow the snapshot to include a small interactive video, that would be immensely useful 

mmm ... apng ... 🤔

FlynnOConnell:
lol I saw mpeg and thought maybe we could make the snapshot a movie... I guess that would then be several snapshots

almarklein:
Well, images can be embedded like ``, and I'm pretty sure that will work with apng (animated png), but for mpeg/mov we'd need a file, so that'd be much harder.

Maybe `.snapshot(1)` will record for 1 second 😄. I think it's possible, but would require some interesting tricks to buildup an apng during rendering ...

almarklein:
Snapshot logic in jupyter_rfb was simplified and the same mechanism copied here.

Issue #102, opened 2025-07-16
New Backend: libavfilter virtual device (ffmpeg/ffplay)?


Vipitis:
Currently more of a goal, I don't have a working POC yet. 

While looking into solutions for recording directly to video, I wanted to still have a GUI for interactivity (for https://github.com/pygfx/shadertoy/pull/55). One thing I read in the documentation is https://www.ffmpeg.org/ffmpeg-devices.html#lavfi if it works could open up some potential. Like writing frames to a encoder (which could also be the streamed remotely).

There is a GUI to this to via `ffplay` but I haven't had the time to look into potential python bindings.
Worst case it's a worse offscreencanvas, which wouldn't make sense to upstream.

Opening this issue to invite suggestions, I hope to have a prototype sometime later this week.

Vipitis:
I am part way to a proof of concept... managed to stream locally to my laptop (albeit awful delay and quality).


![Image](https://github.com/user-attachments/assets/fdb2ab91-9145-4126-9724-0d659123b6d0)

Vipitis:
few evenings of reading through ffmpeg documentation later and I can now say: it totally doesn't work the way I thought it might.
What I ended up is more or less just a sheduled offscreen canvas with the file encoding ontop. I also have ffplay to show what is being recorded via `subprocess.PIPE`. but that is like way worse than a real screen-gui and not interactive.
For GLFW specifically I had something that works, but it's not a good approach. I think to make this work you would likely just have a texture download method for gui canvases and encode that frame on the CPU. Sharing buffers didn't seem possible from my reading of the source.

I tried to keep the commits as sorta useful steps if anyone wants to have a look at it (PR linked above) - but I don't think it's worth upstream something like that... maybe as an example? otherwise feel free to close

hmaarrfk:
I think it is useful to upload your sample code to a zip here

PR #202, opened 2026-03-26
Add http backend

almarklein:
A backend that runs a webserver that you can connect to with your browser. Multiple clients are supported (eventually).

Using ASGI, so it can run on any ASGI server and we don't have any hard dependencies 🚀 

* [x] ASGI server.
* [x] Websocket connection.
* [x] Implement JS to be a AFM (anywidget front-end module) host. Or JS to control the renderview directly.
* [x] Python backend module.
* [x] Support for multiple simultaneous clients.
* [x] Support for including this in a larger web app, e.g. FastAPI.
* [x] Allow customizing the html and e.g. add a logo.
* [x] docs
* [x] Cleanup

----

WIP AFM host code, saving here for reference, but I think I'll skip the afm:
```js

        import {default as afm_factory} from './renderview-afm.js'
        const afm = afm_factory();

        // Open websocket
        const ws = new WebSocket("ws://" + location.host + "/ws");

        ws.onopen = (e) => {


        };
        ws.onerror = (e) =?
        ws.onmessage = (e) => {
            let msg = e.data
            if msg.
        }
        ws.onclose = (e) => {
        }

        const callbacks = {}
        const store = {"css_height"}
        let storeChanges = {}

        class AfmModel {
            on (eventName, callback) {
                callbacks[eventName] |= []
                callbacks[eventName].push(callback)
            }
            get (key) {
                return store[key]
            }
            set (key, value ) {
                store[key] = value
                storeChanges[key] = value
            }
            save_changes () {
                ws.send({'type': 'storeChanges', 'data': storeChanges})
                storeChanges = {}
            }
            send (ob) {
                ws.send(ob)
            }
        }

        const model = AfmModel();
        const el = document.getElementById('canvas')

        afm.initialize({model});
        afm.render({el})
```

### DIFF
--- a/examples/cube_http.py
+++ b/examples/cube_http.py
@@ -2,10 +2,12 @@
 Cube in the browser
 -------------------
 
-Run a wgpu example with the http backend. Also see web_fastapi.py for
-how to integrate a rendercanvas into a larger web application.
+Run a wgpu example with the http backend. Note that the http backend can be used
+with most examples by simply using ``from rendercanvas.http import RenderCanvas,
+loop``. This example also shows how the web-page can be customized.
 
-This example also shows how the web-page can be customized.
+Also see fastapi_app.py for how to integrate a rendercanvas into a larger web
+application.
 """
 
 # run_example = false
--- a/examples/cube_http.py
+++ b/examples/cube_http.py
@@ -55,5 +55,6 @@ im = np.random.uniform(0, 255, (16, 16, 3)).astype(np.uint8)
 resources["logo.png"] = "image/png", encode_png(im)
 
 
-# the loop.run() of this backend uses uvicorn to start a webserver
-loop.run()
+# The loop.run() of this backend uses uvicorn to start a webserver.
+# The args are optional and default to "localhost" and port 60649
+loop.run("localhost", 8080)
--- a/rendercanvas/core/renderview.js
+++ b/rendercanvas/core/renderview.js
@@ -138,7 +138,7 @@ class BaseRenderView {
     this._lsize = null // cached logical size
     this._wheelThrottle = 20 // to avoid flooding wheel events
     this._moveThrottle = 20 // to avoid flooding move events
-    this._isVisible = false // set by intersection observer
+    this._isVisible = 0 // bitmask: 1->intersected, 2->nonzerosize
 
     this._focusElement = null
     this._abortController = new AbortController()
@@ -173,8 +173,12 @@ class BaseRenderView {
     this.viewElement = null
     this.sizeElement = null
     this.titleElement = null
+    this.butCloseElement = null
     if (this.wrapperElement) {
       this.wrapperElement.innerHTML = ''
+      this.wrapperElement.classList.remove('renderview-wrapper')
+      this.wrapperElement.style.width = ''
+      this.wrapperElement.style.height = ''
       this.wrapperElement = null
     }
     const event = {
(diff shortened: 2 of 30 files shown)


almarklein:
Progress! You can basically replace one line to say `from rendercanvas.http import RenderCanvas, loop`, and run the same code, but now its a webserver 🚀 

image

almarklein:
This is nearly done. Need some cleanup, and resolving some small todos.  The server can take multiple clients, who each see the same rendered image. The stream is currently throttled by the slowest connection. Only one client is active: it determines the size of the canvas, and the controls the mouse etc.

rendercanvas/core/renderview.js
```diff
@@ -917,7 +917,7 @@ class BaseRenderView {
       }
       // Prevent the text box from growing
       if (!ev.isComposing) {
-        this.focus_el.value = ''
+        this._focusElement.value = ''
```
almarklein, line 920: This fixes an upstream bug.

almarklein:
A review would be great. Otherwise I'll just merge this in a few days.

rendercanvas/core/renderview.js
Vipitis:
```suggestion
        this._focusElement.focus({ preventScroll: true, focusVisible: false })
```

another potential upstream bug?

rendercanvas/http.py
```diff
@@ -0,0 +1,604 @@
...
+    pass
+
+
+class HttpRenderCanvas(BaseRenderCanvas):
+    """A remote canvas that is served over http and viewed in a browser.
+
+    It is assumed that there is exactly one canvas per connected client.
+    Multiple clients can simultaneously connect to the server. They will be served
+    the same stream of images. There is one "active" client, which determines
+    the pase of rendering. Events from the passive clients are ignored.
```
Vipitis:
```suggestion
    the pace of rendering. Events from the passive clients are ignored.
```

examples/cube_http.py
```diff
@@ -0,0 +1,59 @@
...
+resources["index.html"] = "text/html", html
+
+
+# You can also add new resources, like images or even extra web pages.
+im = np.random.uniform(0, 255, (16, 16, 3)).astype(np.uint8)
+resources["logo.png"] = "image/png", encode_png(im)
+
+
+# the loop.run() of this backend uses uvicorn to start a webserver
+loop.run()
```
Vipitis: maybe document that `loop.run(host="address", port:1234)` takes two kwargs. Potentially the base loop `.run` should consume kwargs for portability?

Vipitis (commented):
I have no webdev experience, so I can't give any input on the core changes. 

Managed to try it even with multiple clients. I am not sure what the timeout is, but if the active client tabs away it basically freezes the rendering - but it remains the active client.

Should there be some notice about the bidirectional nature by default? Because you might receive events from untrusted clients that could somehow be malicious (I am not sure how this would be exploited beyond a denial of service, by like requesting really large resolutions for example? or maybe a close event? - people will find a way).

finally an idea: what might be possible when running multiple backends at the same time? For example one "presenter" doing it locally while others observe via http (or like a frame server that encodes a video file to save/stream it)

rendercanvas/core/renderview.js
```diff
@@ -596,7 +596,7 @@ class BaseRenderView {
     viewElement.addEventListener('pointerdown', (ev) => {
       // When pointer is down, set focus to the focus-element.
       if (!LOOKS_LIKE_MOBILE) {
-        this._focusElement.focus({ preventScroll: true, focusVisble: false })
+        this._focusElement.focus({ preventScroll: true, focusVisible: false })
```
almarklein, line 599: Also fixed upstream: https://github.com/pygfx/renderview/pull/23. @Vipitis  How the hell did you spot that??

Vipitis: I run [typos](https://crates.io/crates/typos-cli) and this was flagged (among a bunch of other pieces around rendercanvas commenst and docs which I didn't care about at the time).

examples/cube_http.py
almarklein: Good idea. Though I don't want to swallow args. If a user wants it to run on multiple backends, they can just omit the args.

almarklein:
> but if the active client tabs away it basically freezes the rendering - but it remains the active client.

Right, if the active client tabs away, the browser automatically scales down the animation loop, thereby slowing the animation. And since the active client determines the pace, the other clients get 1 fps or so too. I will add a note in the docs.

> Should there be some notice about the bidirectional nature by default? Because you might receive events from untrusted clients that could somehow be malicious

Authorization is the responsibility of the user. In practice, each user will have been authenticated, or connect via a magic link.

That said, making another client the active one once the active client drops may not be acceptable behavior in many use-cases. I consider this a first experimental version; we'll have to flesh out details like that if/when this gets used in real-world scenarios. I will add some notes about this too.

> what might be possible when running multiple backends at the same time? For example one "presenter" doing it locally while others observe via http 

Although it sounds like an interesting use-case, I suspect that most ppl would simply open a Google Meet 😅 

I recon you're also thinking of #102, as in having a primary canvas and then a secondary to stream / watch along.

PR #210, opened 2026-04-28
Consistent GLFW backend detection on hybrid Wayland/X11 systems

mortacious:
On linux systems where the glfw library supports  both X11 and Wayland (e.g. the universal `libglfw3` package), `get_x11_window` and `get_x11_display` are present as attributes even when glfw is actually running on the Wayland backend. The old detection logic relied solely on `hasattr(glfw, "get_x11_window")`, which incorrectly selected the X11 path and caused a `TypeError: ...` crash because `get_x11_display()` returned None.

This replaces the attribute check with a runtime probe of the actual display handles: `get_wayland_display()` is called first; `get_x11_display()` is only called when Wayland returns None. Both calls are wrapped in `warnings.catch_warnings(ignore, GLFWError)` to suppress "X11: Platform not initialized" / "Wayland: Platform not initialized" warnings from glfw.

### DIFF
--- a/rendercanvas/core/coreutils.py
+++ b/rendercanvas/core/coreutils.py
@@ -345,9 +345,6 @@ def close_agen(agen):
 SYSTEM_IS_WAYLAND = "wayland" in os.getenv("XDG_SESSION_TYPE", "").lower()
 
 if sys.platform.startswith("linux") and SYSTEM_IS_WAYLAND:
-    # Force glfw to use X11. Note that this does not work if glfw is already imported.
-    if "glfw" not in sys.modules:
-        os.environ["PYGLFW_LIBRARY_VARIANT"] = "x11"
     # Force Qt to use X11. Qt is more flexible - it ok if e.g. PySide6 is already imported.
     os.environ["QT_QPA_PLATFORM"] = "xcb"
     # Force wx to use X11, probably.
--- a/rendercanvas/core/coreutils.py
+++ b/rendercanvas/core/coreutils.py
@@ -344,7 +344,24 @@ def close_agen(agen):
 
 SYSTEM_IS_WAYLAND = "wayland" in os.getenv("XDG_SESSION_TYPE", "").lower()
 
+
 if sys.platform.startswith("linux") and SYSTEM_IS_WAYLAND:
+    # ----- Tweak glfw
+
+    # If you ``apt install glfw3``, you can force pyglfw to use it like so:
+    #
+    #    os.environ["PYGLFW_LIBRARY"] = "libglfw.so.3"
+
+    # Force glfw to use X11, unless it's too late or if the user seems to explicitly target a library.
+    if (
+        "glfw" not in sys.modules
+        and not os.environ.get("PYGLFW_LIBRARY_VARIANT")
+        and not os.environ.get("PYGLFW_LIBRARY")
+    ):
+        os.environ["PYGLFW_LIBRARY_VARIANT"] = "x11"
+
+    # ----- Tweak Qt
+
     # Force Qt to use X11. Qt is more flexible - it ok if e.g. PySide6 is already imported.
     os.environ["QT_QPA_PLATFORM"] = "xcb"
     # Force wx to use X11, probably.
--- a/rendercanvas/glfw.py
+++ b/rendercanvas/glfw.py
@@ -17,7 +17,7 @@ import glfw
 
 from .base import BaseRenderCanvas, BaseCanvasGroup
 from .asyncio import loop
-from .core.coreutils import SYSTEM_IS_WAYLAND, weakbind, logger
+from .core.coreutils import SYSTEM_IS_WAYLAND, weakbind
 
 
 # Make sure that glfw is new enough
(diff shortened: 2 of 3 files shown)


almarklein (commented):
Thanks! Great PR.

I think it makes sense to drop the `api_is_wayland` flag, and move the logic at the top of the module inside the `get_glfw_present_info()`  function. So we have one clear point where the triage happens. What do you think?

mortacious:
> Thanks! Great PR.
> 
> I think it makes sense to drop the `api_is_wayland` flag, and move the logic at the top of the module inside the `get_glfw_present_info()` function. So we have one clear point where the triage happens. What do you think?

Yes, the import-time checks are actually quite useless now, since they are executed again at runtime anyway.

almarklein (approved):
Awesome, thanks!

rendercanvas/glfw.py
```diff
@@ -135,22 +126,49 @@ def get_glfw_present_info(window):
...
+        if SYSTEM_IS_WAYLAND:
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", glfw.GLFWError)
+                wayland_display = (
+                    glfw.get_wayland_display()
+                    if hasattr(glfw, "get_wayland_display")
+                    else None
+                )
+
+            if wayland_display is None:
```
almarklein: This will error when used with glfw that does not have `get_wayland_display`, i..e uses xwayand, right?

mortacious: Indeed! The current code raises if obtaining a wayland window does not work and SYSTEM_IS_WAYLAND is True and does not attempt to fall back to xwayland. This was unintentionally a bit too aggressive. I've just changed it so that it always tries to fall back to X11 if Wayland doesn't work and also removed the workaround for glfw in coreutils.py that messed around with the glfw initialization.

rendercanvas/glfw.py
```diff
@@ -135,22 +126,44 @@ def get_glfw_present_info(window):
...
 
     elif sys.platform.startswith("linux"):
-        if api_is_wayland:
+        # When the respective platform (X11 or Wayland) is not initialized, glfw emits a GLFWError warning and returns None. We treat None as "backend not active" here so that the warning carries no actionable information for the caller.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", glfw.GLFWError)
+            wayland_display = (
+                glfw.get_wayland_display()
+                if hasattr(glfw, "get_wayland_display")
+                else None
```
almarklein, line 135: What happens with a glfw lib that does have `get_wayland_display`, while running on x11?

almarklein:
I'll do some tests when I have access to my Linux box soon.

rendercanvas/glfw.py
mortacious: If `get_wayland_display` fails (either because no wayland display was found or the function does not exist), the code automatically falls back to the x11 code below so it should just work out of the box.

rendercanvas/core/coreutils.py
```diff
@@ -345,9 +345,6 @@ def close_agen(agen):
 SYSTEM_IS_WAYLAND = "wayland" in os.getenv("XDG_SESSION_TYPE", "").lower()
 
 if sys.platform.startswith("linux") and SYSTEM_IS_WAYLAND:
-    # Force glfw to use X11. Note that this does not work if glfw is already imported.
-    if "glfw" not in sys.modules:
-        os.environ["PYGLFW_LIBRARY_VARIANT"] = "x11"
```
almarklein, line 350: This change means that with default `glfw` Python lib installed, on Wayland, the window has no decorations, which would be a regression.

You mentioned the `libglfw3` system package. How do you tell pyGLFW to use that system libary, or does that happen automatically?

mortacious: The glfw package should choose the appropriate library automatically according to its [documentation](https://github.com/FlorianRhiem/pyGLFW), although you can force it by setting `PYGLFW_LIBRARY_VARIANT`. It will also automatically determine which specific .so file to load upon import (on my system, it actually prefers the system library over the locally installed one for some reason). You can also optionally force a specific .so with `PYGLFW_LIBRARY`.

I've just tested it on my system again, and the window shows the correct decorations in both cases. For safety, I forced the .so to the bundled one in `site-packages/glfw/<wayland/x11>/libglfw.so`.

Wayland:
glfw_wayland

X11:
glfw_x11

The missing decoration might be an underlying issue with your Wayland compositor. Decorations on Wayland are a frustrating mess at the moment because windows are responsible for drawing their own decorations (server-side decorations like on X11 are not available unless you force the applications into Xwayland), and each implementation does it differently, especially GNOME. This is also the reason why my Wayland window has a different decoration in the screenshot.

almarklein: I did `apt install libglfw3`, but glfw prefers the packaged lib, because it's a newer version (3.4 vs 3.3.10 for libglfw.so.3).

When I force it to use the system glfw, the decorations are ok, but with the packaged glfw the decorations are not there. This is on Ubuntu 24.04 LTS. I think we'll need to force x11 until the lib from glfw itself works properly on Wayland.

My proposal would be to bring back the setting of PYGLFW_LIBRARY_VARIANT, but only if it's not already set and if PYGLFW_LIBRARY is not set. That way we still fall back, but make it easier for users to disable the fallback.

almarklein: I applied my own suggestion. In summary, if you want to make rendercanvas use the native `libglfw`, you should either set `PYGLFW_LIBRARY` or `PYGLFW_LIBRARY_VARIANT` env vars, or import glfw before rendercanvas does.

almarklein (approved):

PR #213, opened 2026-06-16
Fix HDR support

almarklein:
Fixes this wgpu example: https://github.com/pygfx/wgpu-py/blob/main/examples/offscreen_hdr.py

cc @Vipitis 

* [x] Fix that the returned bitmap download dict had the format hardcoded to 'rgba-u8' instead of using the actual format.
* [x] Fix the conversion of the buffer bytes to a numpy array, for formats other than u8.
* [x] Performance improvement by caching the parsing of the texture format. 
* [x] Add a test to avoid regression. 

----

The wgpu-py example now produces a nice cube:

image


Vipitis (approved):
looks good to me!

Issue #125, opened 2025-11-01
Define how bitmaps with different size than target should be handled


almarklein:
For rendering bitmaps, the given bitmap may be of a different size than the target; the backends take care of the scaling. But should the scaling be nearest-neighbour, linear, or even cubic? And should the user be able to control it? 

Not super urgent, but good to keep in mind and at some point straighten out the differences between backends in this regards.

PR #216, opened 2026-06-25
Add terminal backend

### LINKED ISSUE #207: Backend: Terminal
Would be super-cool to have a backend that renders in the terminal 🤘

And could actually be useful to quickly test something from a terminal session, especially if you're on a remote machine via SSH.

One (simple) approach is to make each char a pixel, and simply set that chars background to the appropriate color. A more sophisticated approach would use ASCII rendering so that a group of pixels can map to a certain character, taking coverage and shape of the chars into account. I saw a super-interesting blog-post or video about that some weeks ago but I cannot find the link right now.

I have some code [here](https://github.com/almarklein/pyterm/tree/main/pyterm/term) from which we could take what's needed; I think we could make it dependency-free.

(not a high priority, mostly making an issue to allow my head to forget the idea for now 😄)

almarklein:
Closes #207 

I'm presenting rendercanvas at Euroscipy in a few weeks, and a terminal backend would be a lovely surprise during a live demo.

Screen Recording 2026-06-26 at 16 25 22

AI disclaimer: I used Claude as lazy way to get the `blessed` calls I need. But I cleaned them up and made them more efficient in most cases.

### DIFF
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -57,6 +57,11 @@ The table below gives an overview of the names in the different ``rendercanvas``
           | ``loop`` (an ``AsyncioLoop``)
         - | Backend when Python is running in the browser,
           | via Pyodide or PyScript.
+    *   - ``terminal``
+        - | ``TerminalRenderCanvas``
+          | ``RenderCanvas`` (alias)
+          | ``loop`` (an ``AsyncioLoop``)
+        - | Render in the terminal (using ``blessed``).
 
 There are also three loop-backends. These are mainly intended for use with the glfw backend:
 
--- a/docs/backends.rst
+++ b/docs/backends.rst
@@ -41,6 +41,11 @@ The table below gives an overview of the names in the different ``rendercanvas``
           | ``RenderCanvas`` (alias)
           | ``loop`` (a ``StubLoop``)
         - | For offscreen rendering.
+    *   - ``terminal``
+        - | ``TerminalRenderCanvas``
+          | ``RenderCanvas`` (alias)
+          | ``loop`` (an ``AsyncioLoop``)
+        - | Render in the terminal.
     *   - ``http``
         - | ``HttpRenderCanvas``
           | ``RenderCanvas`` (alias)
@@ -64,11 +69,7 @@ The table below gives an overview of the names in the different ``rendercanvas``
           | ``loop`` (an ``AsyncioLoop``)
         - | Backend when Python is running in the
           | browser via Pyodide or PyScript.
-    *   - ``terminal``
-        - | ``TerminalRenderCanvas``
-          | ``RenderCanvas`` (alias)
-          | ``loop`` (an ``AsyncioLoop``)
-        - | Render in the terminal (using ``blessed``).
+
 
 There are also three loop-backends. These are mainly intended for use with the glfw backend:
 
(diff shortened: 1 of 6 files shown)


rendercanvas/terminal.py
```diff
@@ -0,0 +1,324 @@
...
+
+loop = TerminalLoop()
+
+
+class TerminalCanvasGroup(BaseCanvasGroup):
+    pass
+
+
+class TerminalRenderCanvas(BaseRenderCanvas):
+    """A canvas that enders in the terminal.
```
Vipitis:
```suggestion
    """A canvas that renders in the terminal.
```

docs/backends.rst
```diff
@@ -57,6 +57,11 @@ The table below gives an overview of the names in the different ``rendercanvas``
           | ``loop`` (an ``AsyncioLoop``)
         - | Backend when Python is running in the browser,
           | via Pyodide or PyScript.
+    *   - ``terminal``
```
Vipitis: If you add a subsection further down it would also show up in the sidebar and the main page.

Vipitis (approved):
First of all this is awesome! I did quickly look at the diff and also tried it with some pygfx and fpl examples. It worked better than expected.
Via ssh the performance was quite poor, but this could have all kinds of reasons.
All of the bitmap examples (drag, noise, snake) had issues but I think that is due to their hardcoded resolution. 

As a dream in the future: if pygfx and fpl render the graphics and gui layer as two different canvas (maybe also for streaming the gui not as pixels, so it could be run the the frontend?) - maybe the gui (and other text) could be rendered as ascii text in the terminal isntead of pixels
Image

And finally, would it make sense to add a small table to the readme. Something like a compatibility matrix to highlight the various backends supported now and what differences might be?

almarklein:
> All of the bitmap examples (drag, noise, snake) had issues but I think that is due to their hardcoded resolution.

Time to address #125; either support some kind of rescaling in the terminal backend, or require the array to match the physical size.

rendercanvas/terminal.py
```diff
@@ -0,0 +1,355 @@
...
+                term.on_color_rgb(*rgb1) + term.color_rgb(*rgb2) + "▄"
+                for rgb1, rgb2 in zip(top_row, bot_row, strict=True)
+            )
+            term_stream.write(term.move_xy(0, y // 2) + line)
+            # Show close button on the first line. Do here rather then at end to avoid flicker.
+            if y == 0:
+                s = f"ratio {self._pixel_ratio:0.4g} -+ ×"  # noqa: RUF001
+                term_stream.write(
+                    term.normal + term.move_xy(img.shape[1] - len(s), 0) + s
+                )
```
hmaarrfk: This was really fun to play with. it would be great to have an option to disable this debug line. 

I was kinda zooming out to the an ungodly small font size and I didn't even know this existed!

docs/backends.rst
almarklein: done

rendercanvas/terminal.py
almarklein: I made a dropdown

almarklein:
I made a "dev-menu-dropdown":

image

image

Now also works with the drag.py and snake examples.

almarklein:
> As a dream in the future: if pygfx and fpl render the graphics and gui layer as two different canvas (maybe also for streaming the gui not as pixels, so it could be run the the frontend?) - maybe the gui (and other text) could be rendered as ascii text in the terminal isntead of pixels

That's interesting, but the terminal backend is probably too niche for a feature like that :)


> Via ssh the performance was quite poor

Have not tested yet myself, but should be much better with #222!

> would it make sense to add a small table to the readme. Something like a compatibility matrix to highlight the various backends supported now and what differences might be?

A list of backends in the readme sounds good! More details like limitations are probably best listed in the doc page on backends.

FlynnOConnell:
Can I +1 this and mark as immediate priority

FlynnOConnell:
Is what I would say if this wasn't already merged

PR #217, opened 2026-06-26
Tweak auto backend selection

almarklein:
* [x] WX is enabled.
* [x] Using ``RENDERCANVAS_BACKEND`` is not restricted to a hardcoded list of backends.
* [x] Select `terminal` backend when it is detected that this is an SSH session.

### DIFF
--- a/rendercanvas/auto.py
+++ b/rendercanvas/auto.py
@@ -18,29 +18,11 @@ from .core.coreutils import (
 from .base import BaseRenderCanvas, BaseLoop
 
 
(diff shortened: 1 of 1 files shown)


Vipitis (approved):
looks good to me, I also tried most of it.

I couldn't get wx in a notebook to work work for me (vscode notebook implementation): it would still import the anywidget canvas and display that in the notebook... but the event loop was now messed up (likely by wx?) where it only updates when you mouse over it.

Issue #183, opened 2026-02-20
The wx canvas hangs on CI with (the now default) bitmap present


almarklein:
For context see https://github.com/pyapp-kit/scenex/pull/42#issuecomment-3928132733.

cc @gselzer 

Windows-only, I believe.

PR #219, opened 2026-06-30
More backend tests

almarklein:
Trying to reproduce #183

* [x] Add more tests to test closing of canvases, that can be re-used by different backends.
* [x] Yay, wx does indeed not clean up properly. 
* [x] Update CI action version.
* [x] Use py 3.14 in CI unless a specific Python version is needed.  
* [x] Fix in encoder test for png (on Win with py314 zlib compression is now better for lower levels??).


almarklein:
I am merging this now, so that this PR is contained to the new tests and test mechanics. Will contribute the fix in a minute.

PR #227, opened 2026-07-06
Refactor canvas and loop closing logic

almarklein:
This started with an attempt to clean up the wx closing mechanics, but ended up being a complete overhaul of the closing mechanics of the base class and all backends.

We have many more backend-specific tests that run on CI now, helping make rendercanvas more stable across the diffetent backends.

## Changes

* [x] All closing flows must go through `BaseRenderCanvas.close()`.
* [x] When the last canvas is closed, the loop is stopped. The base class handles this.
* [x] When the loop is stopped, all canvases are closed.
* [x] We detect loops starting with a little coroutine, but no longer poll canvases being closed with a loop-task.
* [x] We detect loops closing either by 1) the last canvas being closed, or 2) for async loops by the coro to be cancelled.
* [x] Added comprehensive tests for all backends, especially to test stopping and cleanup for GUI backends. All run on CI.
* [x] The wx loop is now a fully functional loop (it was a bit of a special case before this pr).
* [x] Wx canvas are now cleaned up by gc. 
* [x] For all backends, `loop.run()` without any canvases, does one flush of the event queue and then closes.

## Notes for backend authors

* All flows to close a canvas (e.g. pressing the 'cross' of a window) should lead to calling `canvas.close()`.
* The `_rc_close()` method should directly close the window.
* The `_rc_get_closed()` method is removed; the base class tracks the closed state.
* `Loop.run()` without any canvases should do exactly one event loop iteration.


Vipitis:
Yesterday I tried the anywidget backend in marimo and calling `canvas.close()` caused a recursion issue due to event emitting (still worked in Jupyter notebooks...). 

I didn't file an issue yet, but perhaps it's fixed with the changes here. 

E: yes, this fixed the recursion issue... before I could file an issue - awesome!

PR #230, opened 2026-07-11
All backends use nearest interpolation when bitmap is smaller than physical size

almarklein:
Is now also mentioned in the docs. Not fully official spec, but "recommended for backends".

Backends that already did:
  * glfw (via bitmap-wgu adapter)
  * qt
  * pyodide
  * terminal
  
Updated backends:
* wx
* http
* anywidget

----

Some more context.

This does not affect the wgpu context; there the context provides the texture to write to, which precisely matches the canvas' physical size.

The main reason is that when creating simple low-res bitmaps, like the snake example, the canvas shows these big pixels, rather than a smoothed version of it.

### DIFF
--- a/rendercanvas/contexts/_fullscreen.py
+++ b/rendercanvas/contexts/_fullscreen.py
@@ -94,7 +94,9 @@ class FullscreenTexture:
             sample_count=1,
         )
         self._texture_view = self._texture.create_view()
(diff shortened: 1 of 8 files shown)


Vipitis (approved):
Looks go to me. Sounds like a sensible default.

PR #235, opened 2026-07-21
Two fixes for VSCode

### LINKED ISSUE #218: anywidget in vscode: snapshot stays pending when canvas isn't visible
I am on my laptop at the moment, So my screen is only 1080 pixels tall. When I tried the example notebook I noticed something odd:

https://github.com/user-attachments/assets/fc2eebe2-9aab-4f71-baa3-7c341a6fbc5b

if the canvas is not visible, the screenshot remains pending. So if you never let a cell output the canvas, screenshot remains pending. But also if you scroll the screenshot out of the view. I tried to read the code a bit or even debug step it. I saw there was even logic for this exact case. 

https://github.com/pygfx/rendercanvas/blob/e0f67afae31c6cbe9d82e5586db580a66a797fa0/rendercanvas/anywidget.py#L187-L192

But it seems like we never get there in the first place. _maybe_draw never gets reached, because _shedule_maybe_draw never gets reached but _rc_request_draw seems to only get reached when the canvas is running... so I looked into the self.request_draw() situation and that said it sets a flag that only works in "ondemand" mode.

so I tried that update mode and now the snapshots is response, even without a canvas in view. But not if you never let the canvas display in the first case (while a closed canvas is okay...).

So something in the timeline is not working out. Perhaps it would be okay to but a _shedule_maybe_draw at the end of .snapshot?

almarklein:
* [x] Fix that key events did not work, because vscode took away the focus.
* [x] Fix snapshots. Closes #218 
* [x] Minor version bump

### DIFF
--- a/rendercanvas/_version.py
+++ b/rendercanvas/_version.py
@@ -22,7 +22,7 @@ from pathlib import Path
 
 # This is the base version number, to be bumped before each release.
 # The build system detects this definition when building a distribution.
-__version__ = "2.7.0"
+__version__ = "2.7.1"
 
 # Set this to your library name
 project_name = "rendercanvas"
--- a/rendercanvas/anywidget.py
+++ b/rendercanvas/anywidget.py
@@ -127,7 +127,8 @@ class AnywidgetRenderCanvas(BaseRenderCanvas, anywidget.AnyWidget):
             ),
             display_id=True,
         )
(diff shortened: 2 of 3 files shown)


almarklein:
Upstream fix: https://github.com/pygfx/renderview/pull/24

PR #244, opened 2026-07-28
support for tkinter backend. tested on windows 11

Lcbx:
added support for tkinter backend.
imageimage

tested the PR a bunch on my local machine (windows 11)
feedback from X11 / XWayland is welcome
  
Disclaimer : **I used AI**, though I nitpicked it to death for maintainability / respect to the repo's coding style 
~~I would understand if you told me to eat sh** nonetheless~~
CONTRIBUTING.md seems to allow responsible vibe coding. Yay !


almarklein:
Thanks for this! I'm working on an open pr (#227) that changes requires changes to the backends. I'll come back to this pr once that is done.

rendercanvas/core/loop.py
```diff
@@ -439,7 +439,7 @@ def on_interrupt(sig, _frame):
             else:
                 # Setting the signal can raise ValueError if this is not the main thread/interpreter
                 try:
-                    prev_handlers[sig] = signal.signal(signal.SIGINT, on_interrupt)
```
Lcbx, line 442: probably a bug : this sets SIGINT handler twice (instead of SIGINT and SIGTERM)

Lcbx:
FYI this PR is ready for review (and compatible with #227)

rendercanvas/core/loop.py
almarklein: Looks like it, thanks!

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+        width, height = max(1, round(width)), max(1, round(height))
+        if getattr(self.master, "_is_tk_rendercanvas", False):
+            self.master.geometry(f"{width}x{height}")
+        else:
+            self.configure(width=width, height=height)
+        self._size_info.set_physical_size(width, height, 1.0)
+
+    def _rc_close(self):
+        if self._is_closed:
+            return
```
almarklein: No need for is_closed flag anymore. This method is called exactly once.

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+            self.bind(f"<ButtonPress-{button}>", self._on_pointer_down)
+            self.bind(f"<ButtonRelease-{button}>", self._on_pointer_up)
+            self.bind(f"<Double-Button-{button}>", self._on_double_click)
+
+        self._final_canvas_init()
+
+    def _rc_gui_poll(self):
+        try:
+            self.update_idletasks()
+            if not isinstance(self._rc_canvas_group.get_loop(), TkLoop):
```
almarklein: I suspect that the `update_idletasks` can be moved inside the `if` below. That method does not have to be called when tk loop is running, or does it?

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+
+        height, width = rgba.shape[:2]
+        flat = rgba.cast("B") if rgba.c_contiguous else memoryview(rgba.tobytes())
+
+        rgb = bytearray(width * height * 3)
+        rgb[0::3] = flat[0::4]
+        rgb[1::3] = flat[1::4]
+        rgb[2::3] = flat[2::4]
+
+        ppm = f"P6\n{width} {height}\n255\n".encode() + rgb
```
almarklein: This can be optimized further. Numpy can be used, maybe something like:
* `preamble = f"P6\n{width} {height}\n255\n".encode()`
* Allocate flat numpy array that has preamble extra bytes.
* Create a 3d view of the data where the image will be copied to.
* Copy it with `view[:] = rgba[:,:,:3]`
* Pass this to `PhotoImage`, preferably without making a copy. Does it accept a memoryview?

`memoryview` and `ndarray`  can be converted to one another without making a copy. Converting to bytes usually creates a copy.

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+        else:
+            self.itemconfigure(self._image_item, image=photo)
+
+    def _rc_set_logical_size(self, width, height):
+        width, height = max(1, round(width)), max(1, round(height))
+        if getattr(self.master, "_is_tk_rendercanvas", False):
+            self.master.geometry(f"{width}x{height}")
+        else:
+            self.configure(width=width, height=height)
+        self._size_info.set_physical_size(width, height, 1.0)
```
almarklein: This is not the place to update the size info. Just tell the backend to resize, which will very likely call `_on_resize` (but plz check).

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+            self.master.title(title)
+
+    def _rc_set_cursor(self, cursor):
+        try:
+            self.configure(cursor=_CURSOR_MAP[cursor])
+        except tk.TclError:
+            self.configure(cursor="")
+
+    def _on_resize(self, event):
+        self._size_info.set_physical_size(max(0, event.width), max(0, event.height), 1.0)
```
almarklein: Does tk not support pixel ratio?

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+        self.request_draw()
+
+    def _on_destroy(self, event):
+        if event.widget is not self or self._is_closed:
+            return
+        self._is_closed = True
+        self._photo = None
+        self.submit_event({"event_type": "close"})
```
almarklein: Not needed.
```suggestion
```

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+        The sign is reversed because Tk and rendercanvas use opposite conventions.
+        """
+        if event.num in (4, 5):
+            delta = -100 if event.num == 4 else 100
+        else:
+            unit = 120 if sys.platform == "win32" else 1
+            delta = -100 * event.delta / unit
+
+        # horizontal scroll if shift modifier is pressed
+        horizontal = bool(event.state & 0x0001)
```
almarklein: That's up to the application. Backends that support touch can provide horizontal scroll. All other backends should only provide vertical scroll, and applications can decide to use a modifier for horizontal scrolling.

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+                "dy": 0 if horizontal else delta,
+                "x": event.x,
+                "y": event.y,
+                "modifiers": _modifiers(event.state),
+            }
+        )
+
+    def _key_event(self, event_type, event):
+        key = _KEY_MAP.get(event.keysym, event.char or event.keysym)
+        print(key)
```
almarklein: print

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+
+        if self._image_item is None:
+            self._image_item = self.create_image( 0, 0, anchor="nw", image=photo )
+        else:
+            self.itemconfigure(self._image_item, image=photo)
+
+    def _rc_set_logical_size(self, width, height):
+        width, height = max(1, round(width)), max(1, round(height))
+        if getattr(self.master, "_is_tk_rendercanvas", False):
+            self.master.geometry(f"{width}x{height}")
```
almarklein: What does this do? If the widget is embedded in a larger application I would assume setting the logical size does nothing. Or maybe change the widget size if it's in a container that supports resizing.

rendercanvas/tk.py
```diff
@@ -0,0 +1,460 @@
...
+            return
+        widget = self.master if getattr(self.master, "_is_tk_rendercanvas", False) else self
+        try:
+            widget.destroy()
+        except tk.TclError:
+            self._is_closed = True
+
+    def _rc_set_title(self, title):
+        if getattr(self.master, "_is_tk_rendercanvas", False):
+            self.master.title(title)
```
almarklein: If embedded as a widget in a larger application, ignore setting the title. Setting title is responsibility of the application code.

almarklein (commented):
Thanks for this! I made some comments. 

On MacOS this backend is really slow for large resolutions, but there is some room for improvement for the bitmap mode, so this might be improved.

rendercanvas/tk.py
Lcbx: ok, I'll remove any mention / logic depending on _is_closed

rendercanvas/tk.py
Lcbx: fun thing, first draft used numpy
but I asked LLM to use something else since we could not assume user had numpy installed
fine by me

Lcbx: oh and sorry but tkinter does not take memoryviews or ndarrays. tobytes it is

rendercanvas/tk.py
Lcbx: got it

rendercanvas/tk.py
Lcbx: I have something but I'm not willing to bet on it, so I say we keep it simple stupid until someone smarter comes along

rendercanvas/tk.py
Lcbx: so I remove this then

rendercanvas/tk.py
Lcbx: oh. embarrasing

rendercanvas/tk.py
Lcbx: as you guessed : it makes the enclosing widget request that the window gets at least that size ; tha't why we check if It's a rendercanvas. though I'll probably improve this logic in next commits

rendercanvas/tk.py
Lcbx: yeah I'll move that to TkRenderCanvas, makes more sense

Lcbx:
I think this addresses everything
(I answered comments after making the changes asked, that's why it may seem fast)

rendercanvas/tk.py
almarklein: In earlier versions we indeed did not require numpy. Some code in rendercanvas may still unnecessarily avoid numpy.

rendercanvas/tk.py
almarklein: I have a hidpi screen, so I can test it. Let's implement it the proper way. If we don't support pixel ratio, this backend would be close to useless on hidpi screens.

almarklein: You can also post what you have as a comment here, and I can try it out.

rendercanvas/tk.py
almarklein: Common mistake 🤷

rendercanvas/tk.py
```diff
@@ -0,0 +1,454 @@
...
+            self.update_idletasks()
+        except tk.TclError:
+            pass
+
+    def _rc_present_bitmap(self, *, data, format, **kwargs):
+        if format != "rgba-u8":
+            raise ValueError(f"Unsupported bitmap format {format!r}")
+
+        try:
+            rgba = np.asarray(memoryview(data))
```
almarklein: Not a big thing, but just slightly shorter.
```suggestion
            rgba = np.asarray(data)
```

rendercanvas/tk.py
```diff
@@ -0,0 +1,454 @@
...
+            self._image_item = self.create_image(0, 0, anchor="nw", image=photo)
+        else:
+            self.itemconfigure(self._image_item, image=photo)
+
+    def _rc_set_logical_size(self, width, height):
+        width, height = max(1, round(width)), max(1, round(height))
+        if isinstance(self.master, _RenderToplevel):
+            self.master.geometry(f"{width}x{height}")
+        else:
+            self.configure(width=width, height=height)
```
almarklein, line 296: Do you know what configuring the width and height does? I could not find it. Would be good to add that as a comment.

rendercanvas/tk.py
```diff
@@ -0,0 +1,454 @@
...
+                self.close()
+
+        self.root.protocol("WM_DELETE_WINDOW", close)
+
+    # behave like a tk widget
+    def __getattr__(self, name: str):
+        return getattr(self.root, name)
+
+    def set_title(self, title: str) -> None:
+        self.root.title(title)
```
almarklein: The pattern to follow is to implement `_rc_set_title()` on the widget, which then does something like:
```py
        if isinstance(self.master, _RenderToplevel):
            self.master.root.title(title)
```

almarklein (commented):
There are also some linting errors. Easy to fix: run `ruff format` and `ruff lint` to check locally. 

The pixel ratio it a bigger problem. If the reported physical size is incorrect, rendering to screen on a hidpi display will error, because the textures don't match size. I spent some time (with help with both tk docs and Claude) trying to get the pixel ratio from tk, but without luck. It may not be possible.

For bitmap mode it's less of a problem, because tk will simply scale the bitmap. Therefore, a quick solution would be to only support bitmap mode, at least for now.

If we do want to support pixel ratio, we'd have to query the OS directly, which involves a different solution for each OS, which gets ugly especially on x11 vs wayland, so I'd rather not go there for this particular backend.

rendercanvas/tk.py
Lcbx: geometry() asks the window manager to resize the enclosing Toplevel/window to that size.
configure() set the widget requested size

Lcbx:
> The pixel ratio it a bigger problem. If the reported physical size is incorrect, rendering to screen on a hidpi display will error, because the textures don't match size. I spent some time (with help with both tk docs and Claude) trying to get the pixel ratio from tk, but without luck. It may not be possible.

I lifted the logical size computation from [matplotlib tk backend](https://github.com/matplotlib/matplotlib/blob/adecc563eabdfeb103ca338b50d5191b838ae8e9/lib/matplotlib/backends/_backend_tk.py#L263) (license is permissive PSF/BSD-compatible).
If it's good enough for them...

rendercanvas/tk.py
```diff
@@ -0,0 +1,473 @@
...
+        if sys.platform == "win32":
+            # Tk scaling is relative to 72 DPI, Windows scaling to 96 DPI.
+            return round(float(self.tk.call("tk", "scaling")) * 72 / 96, 2)
+
+        elif sys.platform.startswith("linux"):
+            return self.winfo_fpixels("1i") / 96
+
+        else:
+            # Tk/macOS already works in device pixels for this purpose.
+            return 1.0
```
almarklein, line 325: Unfortunately, this is not the case. (And if it was the case, we'd have to scale width/height in several other places.)

But since on MacOS it already operates in bitmap mode, we basically have the fallback I proposed, but only for MacOS. 

Have you been able to test this on a hidpi screen on Windows and/or Linux?

Lcbx: tested it on windows 11 on the tk_app example
(I have only 1 monitor so I have to logout login between display scale changes)

**scaling = 100**
resize event : 876 650
winfo: 1000 650
tk scaling: 1.3339898243886428
1 inch: 96.04726735598227

**scaling = 150**
resize event : 876 650
winfo: 1000 650
tk scaling: 1.332282448377581
1 inch: 95.92433628318584

**code used**
``` python
print("resize event:", event.width, event.height)
print("winfo:",        loop._root.winfo_width(), loop._root.winfo_height())
print("tk scaling:",   loop._root.tk.call("tk", "scaling"))
print("1 inch:",       loop._root.winfo_fpixels("1i"))
```


it seems windows [automatically scales dpi-unaware processes](https://learn.microsoft.com/en-us/windows/win32/api/windef/ne-windef-dpi_awareness#DPI_AWARENESS_UNAWARE)

So behavior is probably fine on windows

PR #245, opened 2026-07-31
Tweak events example

almarklein:
Make it more like the renderview event tester at https://pygfx.org/renderview/tester.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,13 @@ from rendercanvas.auto import RenderCanvas, loop
 canvas = RenderCanvas(update_mode="continuous")
 context = canvas.get_bitmap_context()
 
+
 @canvas.request_draw
 def animate():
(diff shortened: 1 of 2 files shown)


Vipitis:
Hey, my opinion:
These examples will be looked at by people new to rendercanvas as a reference and tutorial. Which includes absolute beginners too.
As such it might look too daunting at first sight compared to the minimal 20 line version it was before. Perhaps a few comments (to explain how easy it is to setup events; and which parts are just to make it fancy) as well as links to the docs/event spec would help.

almarklein:
Good point! This example is indeed mostly meant for testing purposes. Let's prefix it as such, and keep a simpler version similar to what we had before.

almarklein:
I opted for some comments and minor refactoring. There are other examples that better show how to use events in practice.

PR #250, opened 2026-08-09
Fix: Detect the asyncadapter via a thread-local

### LINKED ISSUE #211: Error when using with ipython
Hi! 

I recently updated my `pygfx`, `wgpu` and `rendercanvas` stack. After the update I started running into issues with a little 3D viewer built on top of your tools. Specifically, running it from `ipython` is now throwing errors I hadn't encountered before.

I managed to track the problem to the canvas - `rendercanvas.qt.QRenderCanvas` in my setup. Here is a minimal reproducible example:

```python
>>> %gui qt
>>> from rendercanvas.auto import RenderCanvas
>>> canvas = RenderCanvas()  # <- hitting enter will make the canvas appear
>>>  print("foobar")  # <- entering any further command will lead to an error
foobar

Incompatible awaitable result <Future pending>. Maybe you used asyncio or trio (this does not run on either)?
Incompatible awaitable result <Future pending>. Maybe you used asyncio or trio (this does not run on either)?
```

Versions:
- `rendercanvas == 2.6.3`
- `ipython == 9.13.0`
- `python == 3.11.8`

Downgrading `rendercanvas` to `2.4.2` made the issue go away. I can see you had a major refactor just after that version which is the likely cause of the change in behavior.

I guess my question is: Is this a bug or expected? If the latter, is there anything I can do in my setup to make it work again?

Thanks in advance!

schlegelp:
This PR fixes #211: `Incompatible awaitable result <Future pending>` when trying to use `rendercanvas` in `ipython` with `%gui qt`.

First the disclosure: I used Claude to write the fix.

Then the gist:

Since v2.5.0 (863f2bf, #157) `rendercanvas` detects the running async framework from the global asyncgen hooks instead of `sniffio`. Under `%gui qt` two loops are nested: `prompt_toolkit`'s asyncio loop owns the hooks, and the Qt loop runs inside its input-hook. So when a Qt timer steps an asyncadapter task, detection answers `"asyncio"`, `asyncs.sleep()` yields an asyncio `Future`, and the adapter — which only understands its own protocol — rejects it and cancels the task, including the scheduler.

The fix is to make the adapter mark the task it is stepping in a thread-local, and the detection checks that first: if the adapter is driving the coroutine, it is the `async` lib, no matter what the hooks say.

Loops on real frameworks (`asyncio`, `trio`, and hence jupyter/terminal/http/pyodide) never create adapter tasks, so they should be unaffected. 

Related:`wgpu/_async.py` seems to have a copy of the same hooks-only detection, i.e. would run into the same problem?

### DIFF
--- a/rendercanvas/core/coreutils.py
+++ b/rendercanvas/core/coreutils.py
@@ -14,6 +14,8 @@ import ctypes.util
 from contextlib import contextmanager
 from collections import namedtuple
 
+from ..utils import asyncadapter
+
 
 # %% Constants
 
@@ -325,8 +327,12 @@ def close_agen(agen):
     """Try to sync-close an async generator."""
     closer = agen.aclose()
     try:
-        # If the next thing is a yield, this will raise RuntimeError which we allow to propagate
-        closer.send(None)
+        # We're closing synchronously, so this is not part of the task that we
+        # may be called from; that way our own sleep() and Event() become no-ops
+        # (when no other async lib is active), and the agen can be closed.
+        with asyncadapter.detached():
+            # If the next thing is a yield, this will raise RuntimeError which we allow to propagate
+            closer.send(None)
     except StopIteration:
         pass
     else:
--- a/rendercanvas/core/loop.py
+++ b/rendercanvas/core/loop.py
@@ -532,7 +532,9 @@ class BaseLoop:
         * The subclass is responsible for cancelling remaining tasks in _rc_stop.
         * Return None.
         """
-        task = asyncadapter.Task(self._rc_call_later, async_func(), name)
+        task = asyncadapter.Task(
+            self._rc_call_later, async_func(), name, self.call_soon_threadsafe
+        )
         self.__tasks.add(task)
         task.add_done_callback(self.__tasks.discard)
 
--- a/rendercanvas/utils/asyncadapter.py
+++ b/rendercanvas/utils/asyncadapter.py
@@ -6,6 +6,7 @@ Intended for internal use, but is fully standalone.
 import weakref
 import logging
 import threading
+from contextlib import contextmanager
 
 # Support sniffio for older wgpu releases, and for code that relies on sniffio.
 try:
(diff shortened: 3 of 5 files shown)


schlegelp:
Now that I can use `rendercanvas` in ipython I can also confirm that `wgpu` also the same issue:

```
Present init error
Traceback (most recent call last):
  File "/Users/philipps/Github/rendercanvas/rendercanvas/core/coreutils.py", line 53, in log_exception
    yield
  File "/Users/philipps/Github/rendercanvas/rendercanvas/base.py", line 605, in _draw_and_present
    result = context._rc_present()
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philipps/Github/rendercanvas/rendercanvas/contexts/wgpucontext.py", line 345, in _rc_present
    awaitable = self._downloader.initiate_download(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philipps/Github/rendercanvas/rendercanvas/contexts/wgpucontext.py", line 409, in initiate_download
    self._awaitable.then(self._action.resolve)
  File "/Users/philipps/.pyenv/versions/3.11.8/lib/python3.11/site-packages/wgpu/_async.py", line 344, in then
    raise RuntimeError(
RuntimeError: Cannot use GPUPromise.then() because no running loop could be detected.
```

If/when you are satisfied with the fix here, I'm happy to open an issue + PR over at `wgpu`.

schlegelp:
@almarklein I just noticed your comment in #247:

> - Set `sys.set_asyncgen_hooks` in handlers, not at start of loop. Should fix some ipython related issues.

Maybe you already beat me to it?

almarklein:
> Maybe you already beat me to it?

Not yet, but plan to work on that soon. I'm in the middle of holiday + a move so I'm not moving very fast right now ...

schlegelp:
I figured people were in the middle of holidays :) 

No rush and enjoy the time off 🏖️

schlegelp:
Can I give you a ping about this? :)

PR #252, opened 2026-08-19
Feature request: Add callback submethod for data handling

hmaarrfk:
Would you consider such a change that would allow us to experiment more deeply with custom methods like "np.asarray" and "stried" and even "jpeg".

I think it would be quite helpful to experiment without deeply changing the API.


almarklein:
Yes, this sounds like a good idea!

PR #253, opened 2026-08-22
Ensure utf-8 encoding for anywidget

Vipitis:
closes https://github.com/pygfx/renderview/issues/27

I checked this in the browser and in VSCode and it now displays fine on windows (no longer this windows encoding thing, which was in the css as it loaded).
Not sure if the pyodide one is needed, since pyodide platform encoding localte is proper unicode as far as I know, but I kept it in.
Only python 3.15 has utf-8 on all platforms by default - which is years out to become minimum version.

Had to set `canvas.set_closable(False)` and then `canvas.set_closeable(True)` for it to update... not sure what that is about.


almarklein:
Thanks!

This type of issue is why I tend to prefer reading as bytes and then using `.decode()` (which defaults to utf-8 for a very long time).

PR #254, opened 2026-08-22
Pass the configured view_formats through to the present texture

hmaarrfk:
Allow users to configure the view_formats for the textures.

---------------------------------
From claude: 
`configure()` accepts `view_formats` and validates each against the context capabilities, and then `_get_current_texture()` creates the texture without them — so a view in any declared format fails and the argument silently does nothing. The surface-backed path already forwards them to the surface configuration; this brings the bitmap path in line, one line. Needs pygfx/wgpu-py#832 and is inert until it lands, so the test skips when the backend does not implement `create_texture(view_formats=...)`.

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -296,26 +296,6 @@ def test_wgpu_context_hdr():
     assert np.all(result * 255 == bitmap)
 
 
-def _create_texture_supports_view_formats(device):
-    """Whether the wgpu backend implements create_texture(view_formats=...).
-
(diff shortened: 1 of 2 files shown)


tests/test_context.py
hmaarrfk, line 1: this PR could be merged immediately without these tests, and we could hold these tests until the next wgpu-py release.

almarklein: Maybe add a version check with a skip?

